PR #1, opened 2013-07-03
[WIP] Tests

tarzzz:
This is a work in progress [WiP] pull request to initiate writing tests for the visualization package.
Some things to note:
1. what method should be used for comparing two lists. either by matching each element explicitly.? or is there any specific method

2). Some classes may not have a point_list. ex. Sphere, or Cylinder. So it would be optional arg, in class Shape, and maybe ommited in subclasses **init**().


chrisdembia:
What do you think about testing setter methods as well? Either `shape.set_color('RED')` or `shape.color = 'RED'`.

chrisdembia:
Maybe you can also test invalid user input. See how you might do this here: .

Also, are you sure we want Shape's to have an origin attribute? (1) the origin attribute is only meaningful/useful/not confusing if we specify in the documentaiton how the shape is drawn with respect to the origin. E.g. for a sphere you would say in the documentation 'Sphere is centered at origin'. For a sphere it's obvious, but it's not for a cube; you might choose the origin to be a vertex or the centroid of the cube; those are two valid options. (2) the location of the sphere is set by the translation of the VFrame containing the Shape, and so allowing the user to specify the origin may be redundant and thus unnecessarily complicate the interface.

tarzzz:
So, We can have origin only for specific shapes,
Also what you mentioned earlier regarding naming attributes differently,
like, Cube.points(), instead of Cube.set_point().
I feel set_point seems more obvious, as we can have set_point methods, and
also get_point methods.

A user would be confused what does a Cube.point() does.
Although documentation would tell him, but still on first seeing, it set_
and get_ would be more implicative.

Thanks

On Wed, Jul 3, 2013 at 1:39 PM, Christopher Dembia <notifications@github.com

chrisdembia:
You make a good point. And I think it's fine if you go that way. Be aware though that there seems to be a convention arising with properties.

tarzzz:
I have made some commits with reference to yesterday's talk with Jason.
They are more tuned to what you had mentioned.

tests/test_vframe.py
```diff
@@ -0,0 +1,7 @@
+from sympy.physics.mechanics import *
+#from pydy-viz import VisualizationFrame
+
+def test_visualizattion_frame():
```
chrisdembia: Small typo

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
+from sympy.physics.mechanics import *
+from numpy.testing import assert_allclose
+#from pydy-viz import Shape, Cube, 
+
+def test_shape():
+    point_list = [[1, 0, 0], [2, 3, 1], [-1, 5, -1], [-4, 2, 5]]
```
chrisdembia: It sounds like you guys decided to have the points be lists rather than tuples ([1, 0, 0] rather than (1, 0, 0)). Why is that?

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
...
+    
+    assert cube.name == 'Cube'
+    assert cube.length == 10
+    assert cube.color == 'BLUE'
+    point_list = [[5, 5, 5], [5, 5, -5], \
+                    [5, -5, 5], [-5, -5, -5], \
+                        [-5, 5, 5], [-5, 5, -5], \
+                           [-5, -5, 5], [-5, -5, -5]]
+
+    assert_allclose(cube.point_list,point_list)
```
chrisdembia: space after comma

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
+from sympy.physics.mechanics import *
+from numpy.testing import assert_allclose
+#from pydy-viz import Shape, Cube, 
+
+def test_shape():
+    point_list = [[1, 0, 0], [2, 3, 1], [-1, 5, -1], [-4, 2, 5]]
+    shape = Shape('shape')
+    
+    shape.point_list = point_list
+    shape.color = 'RED'  #abstract method (abc package), a @property
```
chrisdembia: My understanding is that color is a property of Shape, but it is not an abstract method. Abstract methods are those that have a different implementation for each subclass (of Shape). Like, all shapes must have a "do_something()" method, but they each implement it differently. Do we expect different Shape subclasses to handle color differently?

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
...
+    assert cube.color == 'BLUE'
+    point_list = [[5, 5, 5], [5, 5, -5], \
+                    [5, -5, 5], [-5, -5, -5], \
+                        [-5, 5, 5], [-5, 5, -5], \
+                           [-5, -5, 5], [-5, -5, -5]]
+
+    assert_allclose(cube.point_list,point_list)
+    assert_allclose(cube.center, [0, 0, 0])
+    
+    assert isinstance(cube, Shape) == True
```
chrisdembia: Don't need the `== True`.

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
...
+    
+    assert cylinder.name == 'cylinder'
+    assert cylinder.height == 10
+    assert cylinder.radius == 5
+    assert cylinder.color == 'BLACK'
+    assert_allclose(cylinder.center, [0, 0, 0])
+    
+    assert isinstance(cylinder, Shape) == True
+    
+def test_cone():
```
chrisdembia: So three.js defines cylinders in a way that removes the need for a cone class (you can define cones with their cylinder; by setting the top radius to be zero; http://threejs.org/docs/58/#Reference/Extras.Geometries/CylinderGeometry). I don't like how they do that, but maybe we want to consider matching our shapes to their shapes (Geometry)? There doesn't seem to be a need for us to limit what can be drawn in threejs; right now we would not be able to draw a tapered cone, even though threejs allows it. 

This is a very minor point though, and the extra thought may not be worth the small benefit. Let me know your thoughts though.

If we have a user of pydy-viz who is also familiar with threejs, and they see that we allow drawing less shapes than threejs allows, that user might think "well why did they do that?". So maybe we should have an answer to that question; otherwise we should allow equal functionality.

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
...
+def test_cube():
+    cube = Cube('cube')
+    cube.length = 10
+    cube.color = 'BLUE'
+    cube.center = [0, 0, 0]
+    
+    assert cube.name == 'Cube'
+    assert cube.length == 10
+    assert cube.color == 'BLUE'
+    point_list = [[5, 5, 5], [5, 5, -5], \
```
chrisdembia: threejs allows drawing of Cube geometry, so I don't think we need to generate our own points list.
http://threejs.org/docs/58/#Reference/Extras.Geometries/CubeGeometry

chrisdembia:
Great stuff Tarun. A few more things.
1. It's interesting to introduce the 2D shapes such as the circle. I've never thought about doing something like that. I wonder what a cylinder with 0 height would be drawn like. If that would draw a circle, then Circle can be a subclass of  Cylinder, where we internally set the height as 0.0.
2. I think before you were passing in a lot of the attributes into the constructors; how come you've move away from that? I think the tests you have are good, but what do you think about, in addition, testing that the constructors work as well? I imagine some users would prefer using a constructor because it takes less lines than does defining each property separately.
3. I don't see a reason for us to restrict the colors to a few named colors. I definitely think we SHOULD allow using named colors, as I always find that very convenient. But what about allowing specification of an RGB value as well? Or an RGBA value (last element is the alpha value/opacity)? My vote is we specify color as either as a named color, or as RGB, and we have a separate attribute that is the alpha value.
4. So say I want to use a named color, but I don't know which named colors are available? How will this be documented? It's for this reason that I like enum's in C++. Python doesn't have enums, but you could look here for ideas: http://stackoverflow.com/questions/36932/how-can-i-represent-an-enum-in-python. That way, it's right in the code itself (not just in the documentation, which could become outdated) which named colors are available.

For example, you could create a very simple class:

``` python
class Color:
    RED = 1
    GREEN = 2
```

And use it like:

``` python
myShape.color = Color.RED
```

tests/test_shapes.py
```diff
@@ -0,0 +1,148 @@
...
+    mesh_shape = MeshShape('mesh_shape')
+    point_list = [[2, 3, 1], [4, 6, 2], \
+                         [5, 3, 1], [5, 3, 6], \
+                         [2, 8, 4], [7, 4, 1]]
+    mesh_shape.points = point_list
+    mesh_shape.color = 'PINK'
+    mesh_shape.origin = [0, 0, 0]
+    
+    assert mesh_shape.name == 'mesh_shape'
+    assert_allclose(mesh_shape.points,point_list)
```
chrisdembia, line 134: space after comma

tests/test_shapes.py
tarzzz: No, all shapes handle color in same way. I was not totally clear about abstraction at that point. I will remove the comment though .

tests/test_shapes.py
tarzzz: Except  mesh, every basic geometry can be drawn without an explicit point list. So should we limit this to MeshShape class only?

tests/test_shapes.py
tarzzz: I was just focussing on providing some basic shapes for the viz part. there are a few geometries(which are in three.js) which I am ommitting from the Shape subclass. Also although we can consider cone as a three.js cylinder, with one diameter close to zero. But since we are extending the symbolics module(SymPy), wont it would be good on the symbolic viewpoint, to have cone and cylinder differently ?
I am not sure whether this part would be commited to the SymPy later on. But I am assuming that we would want to commit as much as it is acceptable to SymPy, to reside inside SymPy code only, alongwith the rest of mechanics package.

tests/test_vframe.py
tarzzz: Sorry, I didnt understood this one ..

tests/test_shapes.py
chrisdembia: I think so.

tests/test_shapes.py
chrisdembia: Okay that makes sense.

tests/test_vframe.py
chrisdembia: `test_visualization_frame` instead of `test_visualizattion_frame`; there's just an extra `t`.

moorepants:
Tarun, I didn't mean for you to remove the **init** method tests for these classes when I told you about the @property decorator. I think all of the shapes should be completely defined by their **init** method but the user should be able to seemlessly change attributes of the object without get/set methods. For example:

``` python
shape = Cube(length=1.0, color='red', ...)
assert shape.color == 'red'
assert shape.color_hex == "#FF0000"
assert length == 1.0
shape.color = "green"
assert shape.color == 'green'
assert shape.color_hex == '#00FF00'
```

tests/test_shapes.py
tarzzz: Its because :
 1) It is more convenient to perform numpy operations, like comparing and multiplications etc. on list of lists, as they can be readily converted to numpy arrays, as compared to tuples.
 2) A list of list is analogous to a vector3 or vector4 of three.js, and hence no modification would be needed while passing data from python to three.js

moorepants: Chris,

NumPy is going to be a hard dependency beause we will need the linear algebra stuff for transformations and such. I thought these points should be array_like, like in many numpy methods. numpy.array can take a list of lists, list of tuples, tuple of tuples, etc. So any array_like should be accepted.

``` python
numpy.array([[1, 2, 3], [1, 2, 3]])
numpy.array([(1, 2, 3), (1, 2, 3)])
numpy.array(((1, 2, 3), (1, 2, 3)))
```

chrisdembia:
Have either of you seen python packages or modules that have a dash in their name?

moorepants:
module names are underscored, but _package_ names can be anything you want. I haven't seen any convention.

tarzzz:
Regarding colors for shapes, what would be best?
Either a class . like say color.RED or something like strings, like 'RED' with hexa values mapped to a dictionary?

moorepants:
@tarzzz Check out matplotlib's color module for some inspiration: http://matplotlib.org/api/colors_api.html

you can define colors in multiple ways in matplotlib:

'red' common name strings
(1.0, 0.0, 0.0) rgb tuples
'#FF0000' html hex strings
'0.75' shades of gray
etc

matplotlib is bsd style license, so you can use their code if you want

chrisdembia:
Oh interesting; could @tarzzz use that module directly?

tarzzz:
I have put up some basic tests for VIsualizationFrame ..
Please have a look and comment on the same ...

tarzzz:
Also, wont it be an overkill to use matplotlib as a dependency, only for colors, also because there is not much provision of adding grayshade colors in three.js(that i know of). 
Also I think we can provide some basic implementations of colors(string,hex, and rgba), and they can be improved then, after we have the setup working, rather than working on writing a full blown colors module(the one like matplotlib).

chrisdembia:
Yeah it could be overkill, I agree, unless we need it for something else as well already.

tests/test_vframe.py
```diff
@@ -0,0 +1,55 @@
+from sympy.physics.mechanics import *
+#from pydy-viz import VisualizationFrame
+
+def test_visualization_frame():
+	
+	point_list1 = [[2, 3, 1], [4, 6, 2], \
```
chrisdembia, line 6: indentation?

tests/test_vframe.py
```diff
@@ -0,0 +1,55 @@
...
+    
+    #Testing constructor with rframe/point combination
+    I = ReferenceFrame('I')
+    O = Point('O')
+    O.set_vel(I,0)
+    frame1 = VisualizationFrame('frame1', [I, O], shape=mesh_shape1)
+    
+    assert frame1.name == 'frame1'
+    assert frame1.reference_frame == I
+    assert frame1.point == O
```
chrisdembia, line 28: I would call this the `origin` of the visualization frame; it's more specific than just some point in the VFrame.

tests/test_vframe.py
```diff
@@ -0,0 +1,55 @@
...
+    assert frame1.shape == mesh_shape1
```
chrisdembia, line 29: What did we decide about allowing multiple shapes in one VFrame? I don't remember.

chrisdembia:
The interface/tests for VFrame look great. I still feel that, if we expect to use this package/module for visualization of things other than multi-body systems, there should be a separate subclass of VisualizationFrame that takes a RigidBody.

It's good we test creating a VFrame with RigidBody's. BUT I still feel that these VFrame's should be constructed automatically; it doesn't make sense to me that the user has to define all the RigidBody's, and then do the same thing for VFrame's. Seems like duplicate work. Some piece of code should create VFrames and hook them up to RigidBody's, then give the user access to the VFrame's to add Shape's, etc. Let me know what you think about this. It's not clear to me yet how this would be done.

moorepants:
You don't have to make matplotlib a dependency. You can copy just the
source files you want for the basic color manipulation and include the
license from matplotlib with the source. Then check it all into our repo.

Jason
moorepants.info
+01 530-601-9791

On Sun, Jul 7, 2013 at 1:00 PM, Christopher Dembia <notifications@github.com

chrisdembia:
Ahh open source...

tests/test_vframe.py
tarzzz, line 28: Agreed

tests/test_vframe.py
tarzzz, line 29: We dont have multiple shapes.
I am using test for attribute setting method, i.e. that would be used to change that _one_ shape with some other shape. 1 VFrame would definitely take a single shape ..

tarzzz:
Regarding that we have to define a full rbody for the VFrame, it would be something like this ..
1) If a user is already working with rbodies, its simple .  do  .. 
  frame1 =  VFrame('frame1', rbody,shape=some_shape)
2) If a user doesnt wants to deal with rbodies, 
  frame1 = VFrame('frame1',[rframe,origin],shape=some_shape)
Our **init** methods would parse the second arg to check whether user has supplied, 
1) a rbody, 2) a rframe/point tuple,list etc. 3) a rframe/Particle list tuple etc. 

In any case finally all we need is a rframe/point combination, and I feel that it is a necessity(to have a rframe/point) to define a VFrame .

Regarding how to attach rbody data for animation purposes. we would have a generic method.
VFrame.add_simulation _data(), which would attach a JSON file containing simulation data to that frame.
This is when we dont have a rigidbody to provide, and hence _might_ work for other systems.

I would again mention that the rbody in the constructor here is _only_ for providing rframe/point combination, and It is totally optional to use it or else a rframe/point combination explicitly as I did, or a rframe/Particle combination, that I am writing tests for now ..

I hope that clears this bit. .

chrisdembia:
I see what you're saying but I don't think it solves our desire for modularity, if I understand that correctly. By modularity, my understanding is we want people to be able to use this package independently of pydy. In the case that someone is using the package independently of pydy, then they don't know anything about rigidbody's, and none of the classes they use should know anything about rigidbody's (even as an optional argument. imagine reading the docstring and seeing 'argument can be a rigidbody', but not even knowing what that is). But maybe I misunderstand our desire for modularity. I am not too clear on the VFrame.add_simulation_data() / JSON stuff.

tarzzz:
A complete cutoff from pydy would be difficult, as we will have to redesign our VFRame, to function without rframes and points,  moreover by extent of modularity, as I understand, It should be like we are able to see our visualizations without using the pydy module and mechanics methods(i.e. rbodies, Shapes etc.)

That can be achieved if a user is able to output the desired JSON(*in the relevant format) which is input to our Javascripts, by using any other module or even programming language he wants/ or maybe he justs write that JSON and then provide data from somewhere else. 
Anyways depends on what extent of modularity we want.

moorepants:
I think the JS side should be "modular" in the sense that if you provide the right numerical and meta data you'll get to animate things. But the Python side can be less "modular" for general use. We are in a sense writing the plugin for pydy.

tests/test_shapes.py
```diff
@@ -0,0 +1,248 @@
...
+    circle.color = 'GREEN'
+    assert circle.color == 'GREEN'
+
+    circle.center = [0, -3, 5]
+    assert_allclose(circle.center, [0, -3, 5])
+    
+    assert isinstance(circle, Shape)
+
+def test_mesh_shape():
+    point_list = [[2, 3, 1], [4, 6, 2], \
```
moorepants, line 127: Please test passing in numpy arrays and various other array_like things.

tarzzz:
Also regarding the color, we can simply provide a string of color from a standard as defined in matplotlib's color module, and then we can utilize ColorConverter from the matplotlob module to convert those string colors to RGB and RGBA.
All the colors would ultimately be converted into RGB, RGBA before passing to JS.
So there is not much modifications to the tests. 
Also Do I would require to write tests for colors module as welll? If we are taking it from matplotlib and adding to our repo?

moorepants:
If you take the color converter module, you should also take the tests for the color converter module.

moorepants:
Moved to #3.

PR #3, opened 2013-07-09
Tests shapes

tarzzz:
Added tests for all the shapes.


tests/test_shapes.py
```diff
@@ -0,0 +1,334 @@
+from sympy.physics.mechanics import *
+from numpy.testing import assert_allclose
+#from pydy-viz import Shape, Cube, 
+
+def test_cube():
+    cube = Cube('cube', length=10, \
+                         color='BLUE', center=[0,0,0])
```
moorepants: Technically you should have `0.` `0.0` instead of `0`. In the general sense, Python will assume a type of int if you don't have a decimal point. This goes for all the integers you are passing into the various methods for these shapes.

tests/test_shapes.py
```diff
@@ -0,0 +1,334 @@
...
+    assert cube.color == 'RED'
+
+    cube.center = [6, 5, 3]
+    assert_allclose(cube.center, [6, 5, 3])
+    
+    assert isinstance(cube, Shape)
+
+def test_cylinder():
+    cylinder = Cylinder('cylinder', height=10, radius=5, \
+                                    color='BLACK', center=[0,0,0])
```
moorepants: Are you using actually using the geometric center for the base point of these objects? Seems like each shape will have a base point definition in three.js that may or may not be the actual geometric center of the shape. Can you confirm if that is true or not? If it isn't, then using the keyword "center" here may mislead the user.

Also, when a user attaches this to a VizFrame and they also provide a vector to a point, what does the point line up with? Are these center values added to the point vector in the local reference frame of the VizFrame?

tests/test_shapes.py
```diff
@@ -0,0 +1,334 @@
...
+    
+    assert isinstance(circle, Shape)
+
+def test_mesh_shape():
+    point_list = [[2, 3, 1], [4, 6, 2], \
+                         [5, 3, 1], [5, 3, 6], \
+                         [2, 8, 4], [7, 4, 1]]
+    
+    mesh_shape = MeshShape('mesh_shape', points=point_list, \
+                               color='PINK', origin=[0,0,0])
```
moorepants: What's up with the ALL CAPS in the color names? Are you going to call .lower() in the code? Caps seems odd to me.

tests/test_shapes.py
```diff
@@ -0,0 +1,334 @@
...
+    assert_allclose(circle.center, [0, -3, 5])
+    
+    assert isinstance(circle, Shape)
+
+def test_mesh_shape():
+    point_list = [[2, 3, 1], [4, 6, 2], \
+                         [5, 3, 1], [5, 3, 6], \
+                         [2, 8, 4], [7, 4, 1]]
+    
+    mesh_shape = MeshShape('mesh_shape', points=point_list, \
```
moorepants: Why not just `Mesh` instead of `MeshShape` ? All of your other's are just `Cube`, `Sphere`, etc.

tests/test_shapes.py
```diff
@@ -0,0 +1,334 @@
+from sympy.physics.mechanics import *
+from numpy.testing import assert_allclose
+#from pydy-viz import Shape, Cube, 
+
+def test_cube():
+    cube = Cube('cube', length=10, \
```
moorepants: We talked about this already, but what will the `name` argument be used for in the down stream code? Is it for the `__str__` and `__rpr__` methods? If so you need some tests for these.

PR #4, opened 2013-07-11
[WIP] Added some tests for VisualizationFrame class

tarzzz:
Added some basic tests for Visualization Frames


tests/test_visualization_frame.py
```diff
@@ -0,0 +1,117 @@
...
+    assert frame1.reference_frame == A
+    
+    frame1.origin = P1
+    assert frame1.origin == P1
+    
+    frame1.shape = mesh_shape2
+    assert frame1.shape == mesh_shape2    
+    
+    transform_matrix = 
+    assert_allclose(frame.transform(I,O),transform_matrix)
```
tarzzz: How to compare symbolic matrices here.?

chrisdembia: Can you get a string representation of the symbolic equation, from `__str__()` or `__repr__()`?

moorepants: Take a look at the tests in mechanics in test_kane\* where they have examples of this.

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
...
+                sin(p[0])*sin(p[2]) - sin(p[1])*cos(p[0])*cos(p[2]), \
+                10], [-sin(p[2])*cos(p[1]), \
+                   -sin(p[0])*sin(p[1])*sin(p[2]) + \
+                          cos(p[0])*cos(p[2]), \
+                          sin(p[0])*cos(p[2]) + \
+                  sin(p[1])*sin(p[2])*cos(p[0]),10], \
+                   [sin(p[1]), -sin(p[0])*cos(p[1]), \
+                                   cos(p[0])*cos(p[1]),10], \
+                                    [0, 0, 0, 1]]
+
```
tarzzz: @moorepants , Is it Ok to write it here? and in this manner?
It feels a little cumbersome.

moorepants: Yes, that is typical for symbolic tests. You could generate the symbolic matrix in code here, which is independent of your actual implementation code. But this is good. You can clean it up too:

``` python
transformation_matrix = [
    [cos(p[1])*cos(p[2]), sin(p[0])*sin(p[1])*cos(p[2]) + sin(p[2])*cos(p[0]),
        sin(p[0])*sin(p[2]) - sin(p[1])*cos(p[0])*cos(p[2]), 10], 
    [-sin(p[2])*cos(p[1]), -sin(p[0])*sin(p[1])*sin(p[2]) +  cos(p[0])*cos(p[2]), 
        sin(p[0])*cos(p[2]) + sin(p[1])*sin(p[2])*cos(p[0]),10],
    [sin(p[1]), -sin(p[0])*cos(p[1]), cos(p[0])*cos(p[1]),10], [0, 0, 0, 1]] 
```

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
+from sympy.physics.mechanics import *
+from sympy import symbols, sin, cos
+
+#from pydy-viz import VisualizationFrame
```
moorepants: module names should follow pep8 an dhave underscores `pydy_viz` or `pydyviz`

tests/test_visualization_frame.py
moorepants: All of this initialization code is a good use case for a test class.

moorepants: I'd also recommend making a more fleshed out problem. You could use a single conical pendulum or your 3 -link conical pendulum or any other simple problem that is 3D so that you really are able to test this class fully.

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
...
+                   -sin(p[0])*sin(p[1])*sin(p[2]) + \
+                          cos(p[0])*cos(p[2]), \
+                          sin(p[0])*cos(p[2]) + \
+                  sin(p[1])*sin(p[2])*cos(p[0]),10], \
+                   [sin(p[1]), -sin(p[0])*cos(p[1]), \
+                                   cos(p[0])*cos(p[1]),10], \
+                                    [0, 0, 0, 1]]
+   
+def test_vframe_with_rframe():
+    frame1 = VisualizationFrame('frame1', [I, O], shape=mesh_shape1)
```
moorepants: I think the api would be cleanerif you make use of *args and *kwargs in useful ways. I never liked having to pass in tuples of information (i.e. `(I, 0)`, btw technically that should be a tuple and not a list). You could imagine all of these working:

``` python
frame1 = VisualizationFrame('frame1', I, O, shape=mesh_shape1)
frame1 = VisualizationFrame('frame1', rigid_body, shape=mesh_shape1)
frame1 = VisualizationFrame('frame1', I, particle, shape=mesh_shape1)
```

The `__init__` method could be "smart" enough to know what it is getting.

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
...
+    
+    assert frame1.name == 'frame1'
```
moorepants: Once again, why is there a name attribute? How will it be used? These names are useful for symby expressions but I'm not sure what they will be used for in this code.

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
...
+    assert frame1.reference_frame == I
+    assert frame1.origin == O
+    assert frame1.shape == mesh_shape1
```
moorepants: I don't think you've defined the `__eq__`, `__ne__` methods in your shape class. How do you check if shapes are equal?

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
...
+    frame1.reference_frame = A
+    assert frame1.reference_frame == A
+    
+    frame1.origin = P1
+    assert frame1.origin == P1
+    
+    frame1.shape = mesh_shape2
+    assert frame1.shape == mesh_shape2    
+    
+    assert frame1.transform(I, O).tolist() == transformation_matrix
```
moorepants: Will this assertion work? What are the best ways to compare symbolic matrices in sympy?

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,123 @@
...
+    frame2.shape = mesh_shape2
+    assert frame2.shape == mesh_shape2    
+
+    frame2.reference_frame = A
+    frame2.origin = P1
+    assert frame2.transform(I, O).tolist() == transformation_matrix
+    
+def test_vframe_with_particle():
+    
+    frame3 = VisualizationFrame('frame3', [particle1, A], shape=mesh_shape1)
```
moorepants: `[particle, A]` should technically be a tuple `(particle, A)`

tests/tests.py
```diff
@@ -0,0 +1,139 @@
+from sympy.physics.mechanics import *
+from sympy import sin, cos, symbols
+#from pydy-viz import *
+
+class test_visualizationframe_scene(object):
```
moorepants: class names should be camel case `TestVisualizationFrameScene`

tests/test_visualization_frame.py
tarzzz: well the basic purpose is to provide the **str** and **repr** atrribs for Shape as well as Vframe,
But I was also thinking maybe we can use them in the JS side to tag the objects with their name,
I am not sure, whether it is implemented, But I think that THREE.TextGeometry is for this stuff only.
I mean each shape object can have a name tagged to it in Visualization ..

tests/test_visualization_frame.py
tarzzz: Yes good point. definitely agree.
Shapes can be equal if they have same name, type(Cube,Cylinder etc.) and same dimensions, colors ..

tarzzz: Also what does **ne** stands for?

tests/test_visualization_frame.py
tarzzz: Well I tried it in my interpreter, in the illustrative example. seems to work,,
actually tolist() converts them into simple list of lists, and transformation_matrix is defined as lists of lists too.
On comparing two lists, python will run down to each sympy symbol(in each list) for comparisions, which have their **eq** defined ..

tests/test_visualization_frame.py
chrisdembia: ne is "not equal". Jason, I think Tarun might have meant `shape is mesh_shape1`, to ensure it's the same actual object.

tests/test_visualization_frame.py
chrisdembia: I am in favor of having names for most objects; I think people often find a use for it down the line (similar to what Tarun mentioned, maybe you'll have a GUI like Adam Leeper's where you'll want to be able to edit shapes, in which case you'll want to know their names). It's also useful for debug output.

tests/test_visualization_frame.py
chrisdembia: I like what @moorepants said above though, of just having these be multiple arguments.

tests/test_visualization_frame.py
moorepants: Should the name be an optional keyword argument then? It is very clear to me why you do:

``` python
x = Symbol('x')
```

and not

``` python
x = Symbol()
```

in sympy. And in the ReferenceFrame we use the name for printing purposes:

``` python
a = ReferenceFrame('a')
print(1 * a.x + 2 * a.y)
1 * a.x + 2 a.y
a = ReferenceFrame('boogie')
print(1 * a.x + 2 * a.y)
1 * boogie.x + 2 * boogie.y
```

But I'm not seeing an immediate use case. But optionally displaying a text name in the visualization of the shapes that are moving around may be useful. I just don't want to add in extra stuff to type if it is never used.

tests/test_visualization_frame.py
moorepants: yes, using "is" is better here

tests/test_visualization_frame.py
moorepants: ok, just checking

tests/test_visualization_frame.py
chrisdembia: My opinion is that it should be required. I think if it is optional then the backwards-compatibility of using names later on is as difficult as if names were not an attribute in the first place.

I do agree that the user shouldn't need to type it; I don't think the user should even need to create the VisualizaitonFrame's on their own. This gets back to the discussion of the scene automatically generating a VFrame for each rigid body given to the scene; in this case, the Scene could automatically take care of naming the VFrame's according to its rigid body.

I don't think it's the end of the world though if names are not included. One option is, if the name is omitted, to set the VFrame's name to that of its rigid body.

moorepants: I'm generally fine with having the name attribute. I just wanted to make sure that there was at least one reason for having it, I was more or less playing devil's advocate.

But on the auto generation of v frames. I think there is this general case where we want to automatically generate a shape for every rigid body and particle that is in the system. But what about these cases:
1. You don't want to visualize all of the rigid bodies (i.e. make some invisible for clearer viz of the important ones)
2. You need to visualize particles (they need a reference frame attached to them in correct orientation for whatever shape is used)
3. You want to attach a shape to a mass/inertia-less reference frame (i.e. massless rod in a pendulum)
4. What about visualizing funny common things like springs/dampers that aren't rigid bodies but we'd like to put something in place of our force definitions
5. Visualizing general vectors with 3D arrows (like a force or velocity)

Seems to me that there is a lot of detail that you may want to control specifically for visualizations. But I'm all for having some class/function/etc that can auto-generate v frames for all bodies/particles in a system. And maybe this should be a default way to do things. But we should also be able to easily manipulate the fine grained details of what VizFrames we want.

moorepants: @tarzzz Is there an easy way to make an animated spring visualzation with three.js?

chrisdembia: @moorepants So how about a method on Scene that automatically generates a VFrame for each rigid body; it's optional to construct the scene this way:

``` python
scene = Scene('myAwesomeScene')
scene.from_mutibody_system([rb1, rb2]) # rb1's name is 'rb1', rb2's name is 'rb2'
```

creates:

``` python
scene.visualization_frames['rb1']
scene.visualization_frames['rb2']
```

I don't think `from_multibody_system` is the best name for this method.

Then the more manual way of doing this could be:

``` python
scene.add_visualization_frame(VisualizationFrame(rb1))
```

chrisdembia: @moorepants As for the requirement for names: what if the name is an optional argument for VFrames; it's default value is `unnamed`. But say then a `Scene` requires all its VFrame's to have unique names; so when a VFrame is added to it and its name is `unnamed`, Scene changes its name to `vframe1`, etc; it has a count of how many vframes there are and ensures each VFrame has a unique name?

moorepants: Both of these sound good to me. I like the flexibility.

tarzzz: Regarding names for Vframe, I feel we can use the 'unnamed' approach.

As far as vframe generation using scene object. I am a little confused on the approach @fitze mentioned.
It is just that we define a vframe with a shape. I have been trying to keep it super-general, not knowing even if rbodies exist or not.
Just give the numerical data to a vframe, a shape and it gets done. 

But even If we want to generate vframes from Scene automatically, how are we going to attach shapes to a particular vframe.? I am not sure on that one.

Also It will complicate things more If we are thinking of the child-frames for a VFrame type thing.
I am also not sure whether we should keep nested frames,(i.e. one frame child to another vframe).

tarzzz: @moorepants I was looking for visualizing spring, in three.js, but couldnt find one so far.

moorepants: @fitze How would the auto generation of vframes incorporate choosing particular shapes for each rigid body?

moorepants: @tarzzz maybe ask on the three.js email list if drawing a spring can be done.

tests/test_visualization_frame.py
```diff
@@ -0,0 +1,186 @@
...
+        assert self.frame3 == self.frame3_
+        
+        assert self.frame1 != self.frame2
+        assert self.frame1 != self.frame3
+        
+        assert self.frame2 != self.frame1
+        assert self.frame2 != self.frame3
+        
+        assert self.frame3 != self.frame1
+        assert self.frame3 != self.frame2
```
moorepants: I'm not sure it is a good idea to make all these. Isn't "is" what you really want? For example, if you add these. What does `>`, `>=`, `<`, and `<=` do?

tarzzz: should I remove this test method altogether, and replace the others with  'is' ?
like 
'''
assert self.frame2.shape == self.mesh_shape2
'''
with 
'''
assert self.frame2.shape is self.mesh_shape2

'''
Also I should mention that I have added equality, inequality tests (for Shape) already. If you want I can remove them too ..
anyways we would need to define **eq** , **ne** for Shape class, right?

moorepants: The question is what are you trying to test here? For numerical objects like integers and floats it is very clear what `==`, `>`, `!=`, etc should do. But for a shape what does if mean? If I ask `shape1 > shape2`, should it do anything? Should it figure out which one has a greater volume? If you don't have a planned use for all of these operators on this class, then I wouldn't include the code. If all you want to do is to check to see if two references both point to the exact same object, then you should use `is`.

tests/test_visualization_frame.py
chrisdembia: @tarzzz You may be able to use TubeGeometry in three.js.

@moorepants @tarzzz once the vframes are automatically created, you're free to edit them as you wish; even delete them:

``` python
scene.visualization_frames['rb1'].shape = Cylinder(...)
scene.visualization_frames['rb2'].shape = Sphere(...)
scene.remove_visualization_frame['rb1']
```

I definitely do NOT think that any of the vframes should be hidden from the user. The user should have full access to editing/deleting them. But it would be handy if they could be initially created automatically. I can tell you that if I were using the software, I'd definitely have them created automatically for me if that was an option.

moorepants: cool, i'm fine with having an auto gen method for these

tests/test_visualization_frame.py
tarzzz: i dont have any such thoughts. So I have removed them.

moorepants:
It would be better if you reopened this PR and added your commits from PR #16 instead of having a new one. I'm confused about what code I have and haven't read.

PR #8, opened 2013-07-21
Changed directory structure

tarzzz:


tarzzz:
I have done it earlier so that there are no issues with the code imports and stuff later on

PR #9, opened 2013-07-22
Shape and subclasses

tarzzz:
Added a shape class. It has all basic methods and will act as a Super class for other shapes


pydy_viz/shapes.py
```diff
@@ -0,0 +1,172 @@
...
+        to be used by VisualizationFrame class.
+        """
+        
+    def __str__(self):
+        return 'Shape ' + self._name + ' color:' + self._color
+    
+    def __repr__(self):
+        return 'Shape'    
+
+    def generate_visualization_frame(self, reference_frame, origin,\
```
tarzzz, line 125: @fitze , @moorepants I think it would be more helpful in defining VisualizationFrames, rather than from Scene class.
Here we can directly do ..

A = Shape().generate_visualization_frame(I,O)

aand we have a shape and a vframe in place, also since we dont explicitly need shape class to pass to Scene. we can directly pass A
Scene.add_visualization_frame(A)

chrisdembia, line 125: I don't really understand what you said, but your `generate_visualization_frame` is what's called a factory function (http://en.wikipedia.org/wiki/Factory_method_pattern).

PR #13, opened 2013-07-26
Minor Error correction in tests

tarzzz:


moorepants:
There is no need to make a PR for minor stuff like this. Just merge it into master.

PR #14, opened 2013-07-26
Shape and subclasses

tarzzz:
All classes are finished


tarzzz:
A radius can be an integer. I have included both, because being a litlle lazy like myself, users would prefer ..
shape = Sphere(12)
over 
shape = Sphere(12.0)

Anyways I would want to know what we might run into, if we have both ints and floats?

tarzzz:
As for shapes import \* , I have been unable to figure how to import from parent directory.
Once I do, I will change it to something suitable.

PR #15, opened 2013-07-27
Added some numeric transform tests

tarzzz:


moorepants:
What is the point of opening a PR and immediately merging it?

PR #17, opened 2013-07-28
Visualization frame class

tarzzz:


pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
+from sympy.physics.mechanics import *
```
moorepants: avoid `import *`

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+from shapes import *
+import numpy as np
+from sympy.matrices.expressions import Identity
+from sympy import Dummy, lambdify
+
+class VisualizationFrame(object):
+    """
+    A VisualizationFrame is an object used to draw a particular shape
+    or set of shapes. 
+    It basically provides a ReferenceFrame, and an origin to the
```
moorepants: It associates a ReferenceFrame and Point to a Shape. Remove "basically".

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    particular shape attached to it, and is hence useful for the 
+    visualization and animation of the Shape.
+    It is a partial implementation of Scene-Graph.
```
moorepants: What is Scene-Graph? Remove or add a reference to what "Scene-Graph" is.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
+from sympy.physics.mechanics import *
+from shapes import *
+import numpy as np
+from sympy.matrices.expressions import Identity
+from sympy import Dummy, lambdify
+
+class VisualizationFrame(object):
+    """
+    A VisualizationFrame is an object used to draw a particular shape
+    or set of shapes.
```
moorepants: Just below you say a VFrame can only be attached to one Shape, so how is it a "set of shapes"

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    A VisualizationFrame is an object used to draw a particular shape
+    or set of shapes. 
+    It basically provides a ReferenceFrame, and an origin to the
+    particular shape attached to it, and is hence useful for the 
+    visualization and animation of the Shape.
+    It is a partial implementation of Scene-Graph. 
+    A VisualizationFrame can be attached to only one Shape Object.
+    It can be nested, i.e we can add visualization frames added to 
+    one visualization frames. Nesting can be used to make complex 
+    shape/objects.
```
moorepants: I don't understand nesting from this paragraph.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    It basically provides a ReferenceFrame, and an origin to the
+    particular shape attached to it, and is hence useful for the 
+    visualization and animation of the Shape.
+    It is a partial implementation of Scene-Graph. 
+    A VisualizationFrame can be attached to only one Shape Object.
+    It can be nested, i.e we can add visualization frames added to 
+    one visualization frames. Nesting can be used to make complex 
+    shape/objects. 
+    Alongwith Scene class, It can be used to form a complete 2-node
+    deep scene graph.
```
moorepants: Once again, define "scene graph".

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    reference_frame : ReferenceFrame
+        reference_frame with respect to which all orientations of the 
+        shape takes place, during visualizations/animations.
+    origin : Point
+        point with respect to which all the translations of the shape 
+        takes place, during visualizations/animations.
+       
+        Examples
+        ========
+        TODO: Not sure how to put huge examples here.
```
moorepants: Just type them out.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    name : str
+        Name assigned to VisualizationFrame, default is UnNamed
+    shape : Shape
+        Shape to be attached to the VisualizationFrame
+    reference_frame : ReferenceFrame
+        reference_frame with respect to which all orientations of the 
+        shape takes place, during visualizations/animations.
+    origin : Point
+        point with respect to which all the translations of the shape 
+        takes place, during visualizations/animations.
```
moorepants: This should have some mention of the three ways you can supply the parameters: RigidBody, Particle + Frame, or Frame + Point.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+        ========
+        TODO: Not sure how to put huge examples here.
+        >>> from pydy_viz import Shape
+        >>>
+        >>> s = Shape()
+        >>> s.name
+        'UnNamed'
+    
+    """
+    def __init__(self, *args,**kwargs):
```
moorepants: You need a doc string for **init**. What you have in your class def should go here (or be copied and appended here). Class level docs usually describe the attributes and methods, while each method describes how to use that method.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+        #incremented ..
+
+        #If nothing is supplied ...
+        if not args and not kwargs:
+            print '''WARNING:Initializing VisualizationFrame 
+                              without any ReferenceFrame, Point.
+                              For Visualizations to work, 
+                              You must supply a ReferenceFrame, and a
+                              Point'''
+            self._name = 'UnNamed'
```
moorepants: Why allow initialization without enough args?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+            
+        
+        #If name is supplied ...
+        if i < len(args) and isinstance(args[i],str):
```
moorepants: Please start using Pylint, pyflakes or something to check your code for PEP8 compliance. We are 6 weeks in to the project, these minor details shouldn't be popping up anymore...

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+                i+=2
+                
+            elif isinstance(args[i], Particle) and \
+               isinstance(args[i+1], ReferenceFrame):
+                   
+                self._origin = args[i].get_point()
+                self._reference_frame = args[i+1]
+                i+=2        
+        
+        #ReferenceFrame,Point(or Particle) is supplied ...
```
moorepants: Do you really need to allow both (refframe, point) and (point, refframe) ? I would set it to take these args:

(..., rigidbody)
(..., refframe, particle)
(..., refframe, point)

If you do allow multiple orders for frame, point then you need to test this in your tests.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+                 isinstance(args[i+1], Particle):
+                   
+                self._reference_frame = args[i]
+                self._origin = args[i+1].get_point()
+                i+=2    
+            
+        if i<len(args) and isinstance(args[i], Shape):
+            self._shape = args[i]
+            i+=1    
+
```
moorepants: That was a lot of code for checking the args. If you use duck typing principals you may be able to clean that up significantly. For example:

``` python
try:
    frame = args[0].reference_frame
except AttributeError:
    # isn't a rigid body
    try:
        point = args[0].masscenter
    except AttributeError:
        # must be ref frame and point
        point = args[0]        
        frame = args[1]
    else:
        frame = args[1]
else:
    point = args[0].masscenter
```

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+                self._origin = args[i+1].get_point()
+                i+=2    
+            
+        if i<len(args) and isinstance(args[i], Shape):
+            self._shape = args[i]
+            i+=1    
+        
+        #User might want to use keyword args for assigning stuff ...
+        if kwargs.has_key('name'):
+            if not isinstance(kwargs['name'], str):
```
moorepants: If you think in "duck typing" you can avoid having tons of the `isinstance` calls.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+            
+        if kwargs.has_key('origin'):
+            if not isinstance(kwargs['origin'], \
+                                             Point):
+                raise TypeError('''origin should be a valid
+                                         Point object''')
+            else:
+                self._origin = kwargs['origin']
+            
+        if kwargs.has_key('shape'):
```
moorepants: Why is shape an optional argument. Doesn't every VFrame have to have a shape? I believe shape just be a required arg.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+            if not isinstance(kwargs['shape'], \
+                                             Shape):
+                raise TypeError('''shape should be a valid
+                                         Shape object''')
+            else:
+                self._shape = kwargs['shape']
+                
+        if kwargs.has_key('particle'):
```
moorepants: Why is particle an optional arg? I thought this was passed in args?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+            if not isinstance(kwargs['particle'], \
+                                             Particle):
+                raise TypeError('''particle should be a valid
+                                         Particle object''')
+            else:
+                self._origin = kwargs['particle'].get_point        
+                
+        if kwargs.has_key('rigidbody'):
```
moorepants: Why is rigidbody an optional arg?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+                                             RigidBody):
+                raise TypeError('''rigidbody should be a valid RigidBody''')
+                
+            else:
+                self._reference_frame = kwargs['rigidbody'].get_frame()
+                self._origin = \
+                                   kwargs['rigidbody'].get_masscenter()
+                                   
+        #basic things required, transform matrix and child frames  
+        self._transform = Identity(4).as_mutable()
```
moorepants: I think you should generate the real symbolic transformation matrix right now. The user has given you the ref frame and point, so you should generated it.

``` python
self.transformation_matrix = self._generate_symbolic_transformation_matrix()
```

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+        self.child_frames = []
+        
+    #setting attributes ..
+    def __str__(self):
+        return 'VisualizationFrame ' + self._name
+        
+    def __repr__(self):
```
moorepants, line 117: I'm not sure you need this at all. It is hard to represent a VFrame in the proper **repr** way. So why have it? The default repr will show something like: `<__main__.VisualizationFrame instance at 0x1cb6dd0>`

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    def name(self,new_name):
+        if not isinstance(new_name, str):
+            raise TypeError('''Name should be a str object''')
+        else:
+            self._name = new_name
+    
+    @property
+    def origin(self):
+        """
+        origin attribute of the VisualizationFrame,
```
moorepants: Please follow NumPyDoc recommendations on doc strings: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt and at least start the sentence with a capital letter.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+    def shape(self):
+        """
+        shape attribute of the VisualizationFrame.
+        A shape attached to the visualization frame.
+        NOTE: Only one shape can be attached to a visualization frame.
+        """
+        return self._shape
+    @shape.setter
+    def shape(self,new_shape):
+        if not isinstance(new_shape, Shape):
```
moorepants, line 188: Don't you have this same check above. Why do you need it twice? When you set the Shape in the `__init__` method it will call this property method and you will get the type check.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+            raise TypeError('''shape should be a valid Shape object.''')
```
moorepants, line 189: Use capitals to start sentences.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+        return self._shape
+    @shape.setter
+    def shape(self,new_shape):
+        if not isinstance(new_shape, Shape):
+            raise TypeError('''shape should be a valid Shape object.''')    
+        else:
+            self._shape = new_shape
+               
+    def transform(self, reference_frame, point):
+        if not self.reference_frame:
```
moorepants: Why do you have ref_frame and point as args to this method?

``` python
def transform(self):
```

moorepants: Is there a reason for this check? This bascially says "if self.reference_frame is False or None then raise an error. If I have a str stored as self.reference_frame then this check would pass.

moorepants: This needs a doc string.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+        _point_vector = self.origin.pos_from(point).express(reference_frame)
+
+        self._transform[3, 0] = _point_vector.dot(reference_frame.x)
+        self._transform[3, 1] = _point_vector.dot(reference_frame.y)
+        self._transform[3, 2] = _point_vector.dot(reference_frame.z)
+
+        return self._transform
+        
+    def generate_numeric_transform(self, dynamic, parameters):
+        """Returns a function which returns a transformation matrix given
```
moorepants, line 227: Not a complete doc string. Missing Parameters, Returns, Examples, etc.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+
+        Returns
+        -------
+        transform_matrix : numpy.array, shape(n, 4, 4)
+            A 4 x 4 transformation matrix for each time step.
+
+        """
+        #If states is instance of numpy array, well and good.
+        #else convert it to one:
+        if not isinstance(states,np.ndarray):
```
moorepants: not needed. Wouldn't np.array(states) work fine without an if statement.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+                new[i, :, :] = self.numeric_transform(*args)
+
+        else:
+            args = np.hstack((states, parameters))
+            new = self.numeric_transform(*args)
+
+        self.simulation_matrix = new
+        return self.simulation_matrix
+        
+    def add_child_frames(self, *args):
```
moorepants: I still don't understand what this does. You say that VFrames can have only one shape, so what does this do?

Do you have tests for this?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+
+        """
+        for _frame in args:
+            if not isinstance(_frame, VisualizationFrame):
+                raise TypeError('''frame is not an instance 
+                                   of VisualizationFrame:''' + _frame)    
+            else:
+                self.child_frames.append(_frame)      
+                                 
+    def remove_child_frames(self, *args):
```
moorepants: Why do you need a method for this? Can't I just do `vframe.child_frames.remove(3)`?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,370 @@
...
+        for _frame in args:
+            if not isinstance(_frame, VisualizationFrame):
+                    raise TypeError('''frame is not an instance 
+                                       of VisualizationFrame:''' + _frame)    
+            else:
+                self.child_frames.remove(_frame)          
+                    
+    def generate_simulation_dict(self):
+        """
+        Returns a dictionary of all the info required
```
moorepants, line 289: This doc string needs to explain in great detail what a simulation dictionary is. All keys and values should be clearly explained in the Return section.

pydy_viz/visualization_frame.py
tarzzz: @fitze Hi We discussed to include this method, for nesting vframes, to allow complex shapes for some objects.
But I feel it wont be needed with the new API.
@moorepants I will explain a little bit here.
Why we wanted this method in first place.
We wanted to account for creating complex shapes as a combination of vframes which had simple shapes.
They could be tied together, and could be animated together as one single unit. It required specific arguments(numerical values), for that specific parent frame, which would allow all children to be attached together (in visualization) and move together.

Now even If we dont have this nesting to be allowed, since we are only supplying same set of numerical args (integration values of p's q's and static values of mass etc.) there is no concept left for specific arguments for different visualization frames.
That was certainly a different approach for evaluating numerical transformation matrices, and for the current approach it doesnt seem to stand anywhere.
I would also describe, for clarity, the specific arguments which would have been required in the earlier approach.
Like we have a set of nested frames, for representing a RigidBody, then the specific arguments may have been say mass center values at different times. But now it is not needed.
If you have other ideas please let me know. Otherwise we can discard with nesting of vframes altogether, and I feel it wont affect anything on our project.

moorepants: If we don't have nested vframes then you will have to associate every shape that you want to be attached to one reference frame. Like this:

``` python
vframe1 = VisualizationFrame(ref1, point1, Cube(...))
vframe2 = VisualizationFrame(ref1, point2, Cylinder(...))
vframe3 = VisualizationFrame(ref1, point3, Mesh(...))
vframe4 = VisualizationFrame(ref1, point4, Sphere(...))
```

Note that in general the points for each shape are almost always going to be different for each vframe. This is because you what the shapes to rotate together but you want them fixed in the reference frame with respect to one another and the shapes are unlikely to use the same point (i.e. lie on top of one another).

So it seems to me that having child frames can alleviate two things: 1. typing ref1 over and over again for frames with lots of shapes and 2. not having to explicitly have multiple vframe objects. I could imagine something like:

``` python
# the main vframe
vframe1 = VisualizationFrame(ref1, point1, Cube(...))
vframe1.add(point2, Cylinder(...))
vframe1.add(point3, Mesh(...))
vframe1.add(point4, Sphere(...))
```

Now `vframe1` stores all of these new shapes/points as sub-vframes that all share the same reference frame. When you then generate your json file you can loop through these to pull out the needed data.

@fitze do you have any comments on this? As it was your idea originally.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+import numpy as np
+from sympy.matrices.expressions import Identity
+from sympy import Dummy, lambdify
+
+class VisualizationFrame(object):
+    """
+    A VisualizationFrame is an object used to draw a particular shape.
+    It provides a ReferenceFrame, and an origin to the
+    particular shape attached to it, and is hence useful for the 
+    visualization and animation of the Shape.
```
moorepants: Maybe this is more concise:

A VisualizationFrame represents an object that you want to visualize. It allows you to easily associate a reference frame and a point with a shape.

pydy_viz/visualization_frame.py
moorepants: Move this Parameters section into the `__init__` method.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+    origin : Point
+        point with respect to which all the translations of the shape 
+        takes place, during visualizations/animations.
+       
+    """
+    def __init__(self, *args):
+        """
+        Initialises a VisualizationFrame object.
+        To initialize a visualization frame, we need to supply
+        a name(optional), a reference frame, a point, a shape.
```
moorepants: I think it would be cleared if you say something like:

To initialize a visualization frame a ReferenceFrame, Point, and Shape are required. These ReferenceFrame and Point can be passed provided in three ways:
1. RigidBody: the RigidBody's frame and mass center are used.
2. ReferenceFrame and a Particle: The Particle's Point is used.
3. ReferenceFrame and a Point

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+        >>> frame1 = VisualizationFrame('frame1', I, O, shape)
+        >>> Ixx, Iyy, Izz, mass = symbols('Ixx Iyy Izz mass')
+        >>> i = inertia(I, Ixx, Iyy, Izz)
+        >>> rbody = RigidBody('rbody', O, I, mass, (inertia, O))
+        >>> # Initializing with a rigidbody ..
+        >>> frame2 = VisualizationFrame('frame2', rbody, shape)
+        >>> Pa = Particle('Pa', O, mass)
+        >>> #initializing with Particle, reference_frame ...
+        >>> frame3 = VisualizationFrame('frame3', I, Pa, shape)
+        >>> #These can be changed later too ..
```
moorepants: Everything below here in the example doesn't add much.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+            i += 1
+            
+            #Now next arg can either be a Particle or point
+            try:
+                self._origin = args[i].get_point()
+            except AttributeError:
+                self._origin = args[i]
+                
+        #basic things required, transform matrix and child frames  
+        self._transform = Identity(4).as_mutable()
```
moorepants: I don't see my previous comment, but I believe you should generate the actual transformation matrix here and make it a public attribute.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+        return 'VisualizationFrame ' + self._name
+        
+    def __repr__(self):
+        return 'VisualizationFrame'    
+        
+    @property
+    def name(self):
+        """
+        name attribute of the Visualization Frame
+        """
```
moorepants, line 124: One liners are usually like:

`"""Name of the VisualizationFrame."""`

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+            
+    @property
+    def reference_frame(self):
+        """
+        reference_frame attribute of the VisualizationFrame, 
+        with respect to which all rotational/orientational 
+        transformations take place.
+        """
+        return self._reference_frame
+    @reference_frame.setter
```
moorepants, line 166: pep8, lines between methods

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+        """
+        return self._shape
+    @shape.setter
+    def shape(self, new_shape):
+        if not isinstance(new_shape, Shape):
+            raise TypeError('''shape should be a valid Shape object.''')    
+        else:
+            self._shape = new_shape
+               
+    def transform(self, reference_frame, point):
```
moorepants: I'm not fond of this name. There is the noun: `transformation_matrix` and the verb: `generate_transformation_matrix`. You're attributes should have noun names and your methods should be verbs.

moorepants: Also, why do you pass in the reference_frame and point when you already have them stored on the object.

self.reference_frame and self.origin

pydy_viz/visualization_frame.py
moorepants, line 227: the name of this should probably be something like "generate_numeric_transformation_matrix_function" or "generate_transformation_matrix_function", maybe you can think of something shorter but is a verb and tells what is going on

pydy_viz/visualization_frame.py
moorepants: I'm not sure adding the child frame concept is that useful when you can do something like:

``` python
vframes = [VisualizationFrame(ref1, point, shape) for point, shape in zip(list_of_points, list_of_shapes)]
scene = Scene(..., vframes, ...)
```

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+            
+        if not self.simulation_matrix:
+            #Not sure which error to call here.
+            raise RuntimeError('''Please call the numerical 
+                            transformation methods,
+                           before generating simulation dict ''')
+        else:                   
+            self._data['simulation_matrix'] = self.simulation_matrix.tolist()
+            
+        return self._data
```
moorepants: Why make self._data a private attribute?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+            #Not sure which error to call here.
+            raise RuntimeError('''Please call the numerical 
+                            transformation methods,
+                           before generating simulation dict ''')
+        else:                   
+            self._data['simulation_matrix'] = self.simulation_matrix.tolist()
+            
+        return self._data 
+
+class PerspectiveCamera(VisualizationFrame):
```
moorepants: Why not put the Cameras in a camera.py module? In general, I wouldn't expect to find Camera's in visualization_frame.py.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+        >>> #initializing with reference frame, point
+        >>> camera1 = PerspectiveCamera('frame1', I, O)
+        >>> Ixx, Iyy, Izz, mass = symbols('Ixx Iyy Izz mass')
+        >>> i = inertia(I, Ixx, Iyy, Izz)
+        >>> rbody = RigidBody('rbody', O, I, mass, (inertia, O))
+        >>> # Initializing with a rigidbody ..
+        >>> camera2 = PerspectiveCamera('frame2', rbody)
+        >>> Pa = Particle('Pa', O, mass)
+        >>> #initializing with Particle, reference_frame ...
+        >>> camera3 = PerspectiveCamera('frame3', I, Pa)
```
moorepants: You need some examples showing the new optional args. So far this is the same example as the VisFrame.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,693 @@
...
+        name : name of the PerspectiveCamera
+        fov : Field Of View of the PerspectiveCamera
+        simulation_matrix : a N*4*4 matrix, converted to list, for
+        passing to Javascript for animation purposes, where N is the 
+        number of timesteps for animations.
+        
+        
+        """
+        self._data = {}
+        self._data['name'] = self.name
```
moorepants: Why not call `Super.generate_simulation_dict()` and then just add the new fields. There is no reason to repeat any code that you have in the VizFrame class in this class.

pydy_viz/visualization_frame.py
tarzzz: Actually for transformation matrix generation, we need the reference_frame/point w.r.t which we want transformation_matrix.
if that is equal to the rframe/point with which the vframe is defined, it would be an identity matrix only.

pydy_viz/visualization_frame.py
tarzzz: the reference_frame. point here are different from self.reference_frame, point.
The args here are those rframe/point **with** respect to whom I want transformation matrix.(i.e. generally inertial frame.)

pydy_viz/visualization_frame.py
moorepants: Ok, thank you for that clarification. I forgot.

So why not set self._transform to None or don't set it at all here? Is there a reason to have the identity matrix as a default

pydy_viz/visualization_frame.py
moorepants: Thanks, you are correct. I was mistaken. But please write a doc string here that clearly states that.

pydy_viz/visualization_frame.py
tarzzz: Well If we set , say 'I' as inertial frame(for Scene), and then we set an object in the same plane, It wont need to evaluate matrix, as it is identity anyways, since dcm would be Identity and if origin is also same, 4by4 is also identity,

pydy_viz/visualization_frame.py
tarzzz: Well yes I was myself looking for something similar, but Super.method skipped my mind. I am going to implement same in Shape , as some shapes have only, say radius in addition to Shape object. We would need only two lines in dict generator method then.
Thanks for reminding me of it.

pydy_viz/visualization_frame.py
moorepants: That doesn't really answer my question.

What is the reason for having the symbolic transformation matrix equal to the identify matrix instead of it not existing or being equal to None?

Why define it before the your call your .transform method?

tarzzz: Well actually I dont have any specific reason for defining matrix here.

pydy_viz/visualization_frame.py
tarzzz, line 117: Is there something else we can use for repr?

pydy_viz/visualization_frame.py
tarzzz: So If we promote using this type of code statements, then we dont need child frames at all. If you suggest I can completely remove them.

moorepants: Unless @fitze has some comments about these child frames soon, I think it is fine to remove them.

pydy_viz/visualization_frame.py
moorepants, line 117: I'm not sure repr is needed for anything but this seems to say implement **repr** but not necessarily **str**:

http://stackoverflow.com/questions/1436703/difference-between-str-and-repr-in-python

pydy_viz/visualization_frame.py
moorepants: I'd set it to None or not set it at all.

pydy_viz/visualization_frame.py
tarzzz: I dont have any specific reason for making self._data private or public. What would you recommend?
I thought It would help keeping namespace clean.

pydy_viz/visualization_frame.py
moorepants, line 227: These function and attribute names need to be fixed up. The names should clearly say what the method does or what the attribute is without being overly verbose. They should fit the attribute = noun and method = verb guidelines. Right now the names are not very clear. For example "generate_numeric_transform" I would guess this returns a numeric value , but this actually returns a function that returns a numeric value.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+    It allows you to easily associate a reference frame and a point 
+    with a shape.
+    
+    A VisualizationFrame can be attached to only one Shape Object.
+    It can be nested, i.e we can add/remove multiple visualization frames to 
+    one visualization frame. On adding the parent frame to the 
+    Scene object, all the children of the parent visualization frame
+    are also added, and hence can be visualized and animated.
+    
+
```
moorepants: remove blank line

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+    The ReferenceFrame and Point are required to be provided during 
+    initialization. They can be supplied in the form of any one of these:
+    
+    1)reference_frame, point argument.
+    2)a RigidBody argument
+    3)reference_frame, particle argument.
+    
+    In addition to these arguments, A shape argument is also required.
+    
+
```
moorepants: extra blank lines

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        name : str
+            Name assigned to VisualizationFrame, default is UnNamed
+        shape : Shape
+            Shape to be attached to the VisualizationFrame
+        reference_frame : ReferenceFrame
+            reference_frame with respect to which all orientations of the 
+            shape takes place, during visualizations/animations.
+        origin : Point
+            point with respect to which all the translations of the shape 
+            takes place, during visualizations/animations.
```
moorepants, line 52: shouldn't you have an explanation of the other two types of args: rigidbody and particle?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+    
+    1)reference_frame, point argument.
+    2)a RigidBody argument
+    3)reference_frame, particle argument.
+    
+    In addition to these arguments, A shape argument is also required.
+    
+          
+    """
+    def __init__(self, *args):
```
moorepants, line 33: I think this should be:

``` python
def __init__(self, shape, *args, name=None):
```

Where *args can be:

RigidBody
ReferenceFrame,Particle
ReferenceFrame,Point

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        To initialize a visualization frame a ReferenceFrame, 
+        Point, and Shape are required. These ReferenceFrame 
+        and Point can be passed provided in three ways:
+
+        1) RigidBody: the RigidBody's frame and mass center are used.
+        2) ReferenceFrame and a Particle: The Particle's Point is used.
+        3)  ReferenceFrame and a Point
+
+        Parameters
```
moorepants, line 43: Make the parameter list in the same order that you pass them to the `__init__` method.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        ==========
+        name : str
```
moorepants: name : str, optional

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+            Name assigned to VisualizationFrame, default is UnNamed
+        shape : Shape
+            Shape to be attached to the VisualizationFrame
+        reference_frame : ReferenceFrame
+            reference_frame with respect to which all orientations of the
```
moorepants: A reference frame with respect ...

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        ==========
+        name : str
+            Name assigned to VisualizationFrame, default is UnNamed
+        shape : Shape
+            Shape to be attached to the VisualizationFrame
+        reference_frame : ReferenceFrame
+            reference_frame with respect to which all orientations of the 
+            shape takes place, during visualizations/animations.
+        origin : Point
+            point with respect to which all the translations of the shape
```
moorepants: A point with respect...

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        Examples
+        ========
+        >>> from pydy_viz import VisualizationFrame, Shape
+        >>> from sympy.physics.mechanics import \
+                               ReferenceFrame, Point, RigidBody, \
+                                Particle, inertia
+        >>> from sympy import symbols                               
+        >>> I = ReferenceFrame('I')
+        >>> O = Point('O')
+        >>> shape = Shape()
```
moorepants: why doesn't Shape  take any arguments?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        >>> i = inertia(I, Ixx, Iyy, Izz)
+        >>> rbody = RigidBody('rbody', O, I, mass, (inertia, O))
+        >>> # Initializing with a rigidbody ..
+        >>> frame2 = VisualizationFrame('frame2', rbody, shape)
+        >>> Pa = Particle('Pa', O, mass)
+        >>> #initializing with Particle, reference_frame ...
+        >>> frame3 = VisualizationFrame('frame3', I, Pa, shape)
+
+        """
+        #Last arg should be a Shape ..
```
moorepants, line 84: shape is the second arg in your parameter list in the doc string. See my note above for the `__init__` method to solve this.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+            self._shape = args[-1]
+        else:
+            raise TypeError('''Please provide a valid shape object''')    
+        i = 0
+        #If first arg is not str, name the visualization frame 'UnNamed'    
+        if isinstance(args[i], str):
+            self._name = args[i]
+            i += 1
+        else:
+            self._name = 'UnNamed'
```
moorepants, line 95: Why not use None instead of a default name?

You can specify this in the `__init__` method like `__init__(..., name='unnamed')`

What is with the CamelCase?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+    #setting attributes ..
+    def __str__(self):
+        return 'VisualizationFrame ' + self._name
+        
+    def __repr__(self):
+        return 'VisualizationFrame'    
+        
+    @property
+    def name(self):
+    """
```
moorepants: """Returns the name of the visualization frame."""

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+    Name of the VisualizationFrame.
+    """
+        return self._name
+        
+    @name.setter
+    def name(self, new_name):
+        if not isinstance(new_name, str):
```
moorepants, line 133: """Sets the name of the visualization frame."""

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+            raise TypeError('''Name should be a str object''')
+        else:
+            self._name = new_name
+    
+    @property
+    def origin(self):
+        """
+        Origin of the VisualizationFrame,
```
moorepants, line 141: Returns the origin of the visualization frame.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        Origin of the VisualizationFrame, 
+        with respect to which all translational transformations
+        take place.
+        """
+        return self._origin
+        
+    @origin.setter
+    def origin(self, new_origin):
+        if not isinstance(new_origin, Point):
```
moorepants, line 152: """Sets the origin of the visualization frame."""

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+            raise TypeError('''origin should be a valid Point Object''')
+        else:
+            self._origin = new_origin        
+            
+    @property
+    def reference_frame(self):
+        """
+        reference_frame of the VisualizationFrame,
```
moorepants, line 160: """Returns the reference frame of the visualization frame. ..."""

Be consistent with these doc strings.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        return self._shape
+        
+    @shape.setter
+    def shape(self, new_shape):
+        if not isinstance(new_shape, Shape):
+            raise TypeError('''shape should be a valid Shape object.''')    
+        else:
+            self._shape = new_shape
+               
+    def generate_transformation_matrix(self, reference_frame, point):
```
moorepants, line 193: needs a docstring

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        _rotation_matrix = self.reference_frame.dcm(reference_frame)
+        self._transform = Identity(4).as_mutable()
```
moorepants, line 215: I think this should be a public attribute. Would it be better to name it VisualizationFrame.symbolic_transformation_matrix? Or at least .transformation_matrix?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+
+        return self._transform
+        
+    def generate_numeric_transform(self, dynamic, parameters):
+        """Returns a function which returns a transformation matrix given
+        the symbolic states and the symbolic system parameters.
+        
+        Parameters
+        ==========
+        dynamic : list of all the dynamic symbols used in defining the
```
moorepants, line 233: You should probably mention the order of the dynamic symbols and pararmeters matters. They will map to the order of the numerical function call.

moorepants, line 233: Maybe dynamic_variables is clearer than dynamic.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        
+    def generate_numeric_transform(self, dynamic, parameters):
+        """Returns a function which returns a transformation matrix given
+        the symbolic states and the symbolic system parameters.
+        
+        Parameters
+        ==========
+        dynamic : list of all the dynamic symbols used in defining the 
+                  mechanics objects.
+        parameters : list of all symbols used in defining the
```
moorepants, line 235: We could name this `constant_variables` if the above is `dynamic_variables`.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+
+        dummy_symbols = [Dummy() for i in dynamic]
+        dummy_dict = dict(zip(dynamic, dummy_symbols))
+        transform = self._transform.subs(dummy_dict)
+
+        self.numeric_transform = lambdify(dummy_symbols + parameters,
+                                          transform, modules="numpy")
+        
+    def evaluate_numeric_transform(self, states, parameters):
```
moorepants: I like `evaluate_transformation_matrix` better.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        """Returns the numerical transformation matrices for each time step.
+
+        Parameters
+        ----------
+        states : array_like, shape(m,) or shape(n, m)
+            The m states for each n time step.
+        parameters : array_like, shape(p,)
+            The p constant parameters of the system.
```
moorepants, line 261: If we change the above, I'd go with dynamics_values and constant_values here.

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Returns
+        -------
+        transform_matrix : numpy.array, shape(n, 4, 4)
```
moorepants, line 265: you say it returns transformation_matrix but the code returns something call simulation_matrix and it sets the attribute. Should we store this as an attribute or just return it?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+    
+        
+        """
+
+        dummy_symbols = [Dummy() for i in dynamic]
+        dummy_dict = dict(zip(dynamic, dummy_symbols))
+        transform = self._transform.subs(dummy_dict)
+
+        self.numeric_transform = lambdify(dummy_symbols + parameters,
+                                          transform, modules="numpy")
```
moorepants, line 250: Should this return the function?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+                new[i, :, :] = self.numeric_transform(*args)
+
+        else:
+            args = np.hstack((states, parameters))
+            new = self.numeric_transform(*args)
+
+        self.simulation_matrix = new
+        return self.simulation_matrix
+        
+    def generate_simulation_dict(self):
```
moorepants: Wouldn't `generate_visualization_dict` be a more appropriate name?

pydy_viz/visualization_frame.py
```diff
@@ -0,0 +1,280 @@
...
+        self._data['name'] = self.name
+        self._data['shape'] = self.shape.generate_data()  
+        if not self.simulation_matrix:
+            #Not sure which error to call here.
+            raise RuntimeError('''Please call the numerical 
+                            transformation methods,
+                           before generating simulation dict ''')
+        else:                   
+            self._data['simulation_matrix'] = self.simulation_matrix.tolist()
+        return self._data
```
moorepants: I don't see any reason not to make self._data into self.data. This is something the user always wants access too.

pydy_viz/visualization_frame.py
chrisdembia, line 117: I thought repr was for reproducing the object.

pydy_viz/visualization_frame.py
chrisdembia: Shape should be abstract, right? So you shouldn't really create a Shape; just subclasses.

pydy_viz/visualization_frame.py
chrisdembia, line 95: Yeah I think it's fine if the first arg isn't name; it can just be a keyword arg.

pydy_viz/visualization_frame.py
chrisdembia, line 141: what units? in which frame?

pydy_viz/visualization_frame.py
chrisdembia, line 233: state_variables?

pydy_viz/visualization_frame.py
chrisdembia: +1

pydy_viz/visualization_frame.py
moorepants, line 117: Me too, but for complicated objects, that seems hard to do. And the above stack overflow seems to say differently.

pydy_viz/visualization_frame.py
chrisdembia, line 265: Would we need to cache it? Will it be called multiple times before being changed?

pydy_viz/visualization_frame.py
moorepants: So won't this example fail, Tarun?

pydy_viz/visualization_frame.py
chrisdembia: Why can't you just have self._name, self._shape, etc.? Because having self._data makes the serialization easier? Having self.data may be confusing, if self._name etc exists. Does this (self._data) need to be a member at all, or can it just be returned?

pydy_viz/visualization_frame.py
moorepants, line 141: No units and no frame. It is just a sympy.physics.mechanics.Point object.

pydy_viz/visualization_frame.py
moorepants, line 233: state_variables is potentially good too, but it is possible that in the future we have time varying inputs that may pop up here too.

pydy_viz/visualization_frame.py
moorepants, line 265: Maybe. You will call this once per simulation that you pass in.

pydy_viz/visualization_frame.py
moorepants: I think it can just be returned. It doesn't take much to compute it.

pydy_viz/visualization_frame.py
chrisdembia, line 141: Ohh okay; this says nothing of how the point is expressed.

moorepants, line 141: Correct.

pydy_viz/visualization_frame.py
tarzzz: Actually I had a thought of using Shape() as representing a particle, instead of Sphere with a small radius.
It would be easier , and 
It wont fail.
s = Shape()
 will set ...
s.name = 'UnNamed'
and s.color = 'grey'
and can be used for particle. 
I had this though in mind for some time, but I forgot to discuss it.

pydy_viz/visualization_frame.py
tarzzz, line 265: It would be called once, and normally these transformation related functions are/ would be called by Scene's method.

pydy_viz/visualization_frame.py
tarzzz, line 250: Actually it depends on how we want to call transformation methods. I was inclined to use Scene's methods preferably call all the transformation methods.
That is one of the reason for setting the attributes private.Although It wont matter much.

pydy_viz/visualization_frame.py
tarzzz: So it shouldnt be a self. attribute?

pydy_viz/visualization_frame.py
moorepants: I don't think that is a good idea. You should leave Shape as an abstract base class and keep Sphere as a sub class. The user should never call Shape.

pydy_viz/visualization_frame.py
moorepants, line 265: I could imagine keeping all of these methods private and having Scene be the only thing that calls them internally. I'd be fine with that.

pydy_viz/visualization_frame.py
moorepants, line 250: Ok, that seems reasonable, but you need to be consistent. If Scene is going to call the generate_transformation_matrix, then it should also be private. This makes since to do because you pass Scene the global inertial frame and origin point.

pydy_viz/visualization_frame.py
moorepants: Chris and I don't think it should be an attribute.

pydy_viz/visualization_frame.py
tarzzz: Ok Sure, But anyways this docstring ex. wont fail. But should I modify it to something like Cylinder or Sphere?

moorepants: yes.

pydy_viz/visualization_frame.py
tarzzz, line 95: I just kinda liked the CamelCase, assuming that maybe in future if we keep names in the visualization(as text).
It would look impressive.

As for name as keyword arg, I thought this type would be consistent with sympy mechanics classes, 
like RIgidBody('name', ... other stuff) etc.

moorepants, line 95: The CamelCase strings are not impressive looking, trust me. It just looks unprofessional.

We've discussed multiple times now that name is not a required attribute for many of the objects you are creating. The mechanics objects have names for very specific reasons (i.e. it is used for display vectors properly). I believe object names should be optional for everything you've written so far because they have no explicit use and why have to type in a name for every object instantiation if it never gets used anywhere.

chrisdembia, line 95: If possible, it would be good if we made it easy to make use of names 
later on, if we find a use for it down the line.

moorepants, line 95: Would an optional argument for the name solve that?

tarzzz, line 95: Would be fixed in master .. with keyword arg

pydy_viz/visualization_frame.py
tarzzz, line 141: fixed

pydy_viz/visualization_frame.py
tarzzz, line 193: fixed

pydy_viz/visualization_frame.py
tarzzz, line 235: would be fixed in master

pydy_viz/visualization_frame.py
tarzzz, line 250: fixed

pydy_viz/visualization_frame.py
tarzzz, line 261: to be fixed in master

pydy_viz/visualization_frame.py
tarzzz, line 265: fixed

PR #18, opened 2013-08-01
Scene class

tarzzz:


pydy_viz/scene.py
```diff
@@ -0,0 +1,207 @@
...
+        save_to : str
+            path to the file where to write the generated data JSON.
+            the path should be chosen such as to have the write 
+            permissions to the user.
+
+        Examples
+        ========
+        #TODO : Write complete example for initializing a vframe and stuff, 
+               or directly show this method?              
+        """
```
tarzzz, line 222: @moorepants How should I write examples here?

moorepants, line 222: Yes, ideally all methods should have examples of their use.

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
+from sympy.physics.mechanics import ReferenceFrame, Point  
+from visualization_frame import VisualizationFrame
+from camera import PerspectiveCamera
+from server import start_server
+import json
+import re
+
+try:
+    import IPython
+except:
```
moorepants: except ImportError:

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+    A scene needs to be supplied with visualization_frames, Cameras, 
+    and Light objects, as optional arguments.
+    A scene can also be supplied with the height and width of the 
+    browser window where visualization would be displayed.
+    Default is 800 * 800.
+    
+    
+    """
+    
+    def __init_(self, *args, **kwargs):
```
moorepants: Wouldn't this be better:

`def __init__(self, reference_frame, origin, visualization_frames, name=None, width=800, height=600, camera=None)`

Or 

`def __init__(self, reference_frame, origin, *visualization_frames, name=None, width=800, height=600, camera=None)`

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+        It requires a reference frame and a point to be initialized.
+        
+        Parameters:
+        ===========
+        
+        name : str
+        Name of Scene object.
+        
+        reference_frame : ReferenceFrame
+        All the transformations would be carried out with respect
```
moorepants, line 47: Indent after arg or numpydoc doesn't work.

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+        origin : Point
+        All the transformations would be carried out with respect
+        to this point.
+        
+        width : int or float
+        width of the canvas used for visualizations.
+        
+        height : int or float
+        height of the canvas used for visualizations.
+
```
moorepants, line 68: You are missing camera and visframes.

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+            frame.generate_transformation_matrix(self._reference_frame, self._origin)
+            frame.generate_numeric_transform(state_sym, par_sym)
+            frame.evaluate_numeric_transform(states, parameters)
+            self._scene_data['frames'].append(frame.generate_simulation_dict())
+
+     
+        outfile = open(self.saved_json_file)
+        outfile.write(json.dumps(self._scene_data, indent=4, separators=(',', ': '))) 
+        outfile.close()
```
moorepants: It would be nice to seperate this method into two methods. One that generates a json compatible dictionary and the other that writes the file. Then a user can make use of the python dictionary even if they don't need to dump to a json file.

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+            should be respective to the par_sym.
+
+        save_to : str
+            path to the file where to write the generated data JSON.
+            the path should be chosen such as to have the write 
+            permissions to the user.
+
+        """
+        self.saved_json_file = save_to
+        self._scene_data = {}
```
moorepants: I don't see any reason to make scene_data private.

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+        outfile.close()
+
+    def _display_from_ipython(self, json_data=None):
+        #Basic script requirements. ..
+        pass
+            
+
+    def _display_from_interpreter(self, json_data=None):     
+        #Get html file ..
+        _path_to_html = pydy_viz.__file__[:-12] + 'index.html'
```
moorepants, line 236: the os module gives tools to deal with paths in os.path. Like os.path.split and such. That is better than relying on the number of characters in a str.

Where do you plan to store the default index.html file? Maybe there should be a static directory? Like pydy-viz/pydy-viz/static

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+        outfile = open(self.saved_json_file)
+        outfile.write(json.dumps(self._scene_data, indent=4, separators=(',', ': '))) 
+        outfile.close()
+
+    def _display_from_ipython(self, json_data=None):
+        #Basic script requirements. ..
+        pass
+            
+
+    def _display_from_interpreter(self, json_data=None):
```
moorepants, line 234: What happens if you choose the option to not provide any json data? (i.e. why is it optional to provide the data?

pydy_viz/scene.py
```diff
@@ -0,0 +1,257 @@
...
+        
+        """        
+     
+        try:
+            config = get_ipython().config
+            if config['KernelApp']['parent_appname'] == \
+                                                  'ipython-notebook':
+                self._display_from_ipython(json_data)
+                
+        except:
```
moorepants, line 279: make sure to use the exact error that would be raise here like:

except IpythonError:

pydy_viz/server.py
```diff
@@ -0,0 +1,19 @@
+import sys
+import BaseHTTPServer
+from SimpleHTTPServer import SimpleHTTPRequestHandler
+
+
+def create_server(port=8000):
```
moorepants, line 6: Why don't you pass it the directory here that has the index.html file?

pydy_viz/server.py
```diff
@@ -0,0 +1,19 @@
...
+
+    server_address = ('127.0.0.1', port)
+
+    HandlerClass.protocol_version = Protocol
+    httpd = ServerClass(server_address, HandlerClass)
+
+    sa = httpd.socket.getsockname()
+    print 'Your server is running on http://127.0.0.1:8000/ \n \
+           Visit in browser to see your visulization in all its glory'  
+    httpd.serve_forever()
```
moorepants, line 19: How does the user stop the server?

pydy_viz/tests/test_visualization_frame_scene.py
```diff
@@ -143,87 +153,191 @@ def test_vframe_with_particle(self):
...
         
         self.frame4.name = 'frame1_'
         assert self.frame4.name == 'frame1_'
     
     def test_vframe_nesting(self):
         self.frame5 = VisualizationFrame('parent-frame', self.I, \
-                                        self.O, shape=self.mesh_shape1)
+                                        self.O, self.shape1)
+        
+        self.frame5.add_child_frames(self.global_frame1, \
```
moorepants, line 172: I thought you removed these methods?

pydy_viz/tests/test_visualization_frame_scene.py
```diff
@@ -143,87 +153,191 @@ def test_vframe_with_particle(self):
...
+        assert camera.fov == 30
+        
+        camera.near = 10
+        assert camera.near == 10
+        
+        camera.far = 500
+        assert camera.far == 500
+        
+        #Test UnNamed
+        camera1 = PerspectiveCamera(self.I, self.O)
```
moorepants, line 275: Are there no options in three.js for the perspective camera? The ortho camera as near and far, for example.

pydy_viz/scene.py
tarzzz: def **init**(self, reference_frame, origin, visualization_frames, name=None, width=800, height=600, camera=None)

In this way, we can directly assign it to a list, and can use append, remove methods for list.
It would also appear distinct to other arguments, easily readable ..

s = Scene(I, O, [vframe1,vframe2,..], name='sdf', ...)

rather than ..
s = Scene(I, O, vframe1, vframe2, name='sas', ...)

moorepants: Keep in mind that you can always do this:

``` python
frames = (VisualizationFrame(...), VisualizationFrame(...), VisualizationFrame(...))
s = Scene(I, O, *vframes, name='my scene')
```

i.e.

``` python
>>> def f(*args):
...     for arg in args: print(arg)
... 
>>> things = ('one', 'two', 'three')
>>> f(*things)
one
two
three
```

pydy_viz/scene.py
tarzzz: We can rather have save_to = False for not writing to file .., rather than two methods
User can give save_to = False if he doesnt wants to write to file

pydy_viz/scene.py
tarzzz, line 236: Yes It would be incorporated with the JS directory design, or overall "static" dir. design.
Actually I have kept it a little abstract for now. Once we have figured out how JS would be handled in the package, we can modify it then.

pydy_viz/scene.py
moorepants: That doesn't solve the issue I pointed out. I, as a user, may want to get a hold of the scene data python dictionary. It would be nice to have a method that returns this dictionary. Then have a method that writes that dictionary to a json file.

pydy_viz/scene.py
moorepants, line 236: Ok, but you should still be using os.path.

pydy_viz/scene.py
chrisdembia: Just a small comment; I don't think this is an obvious usage though, for 
most people. If we expect this usage, we should provide examples of it.

moorepants: Agreed, it isn't something that novice pythonista's know. But if we have clear documentation then we can go with that.

I personally don't like all the list passing as args in the current mechanics module. It seems very clunky to me. That's why I recommend this method.

chrisdembia: I agree that the passing in of lists is worse.

pydy_viz/server.py
tarzzz, line 19: Its usually, cttrl + C

moorepants, line 19: I'd recommend copying the elegant way ipython notebook shutsdown when you press crtl+C

pydy_viz/scene.py
tarzzz: Well actually since I am using _args and *_kwargs, would it be possible to use  starred args for VFrames.?

moorepants: There are several issues:
1. Clarity of what args and what order they should be given.
2. Clarity in the docstring what to pass in and how.
3. Reducing the amount of arg parsing code in your method.

My recommended solution is to provide explicit args and kwargs in the function call signature, except for the vframes argument which should be *visualization_frames and the user can pass in multple v frames manually or by the *vframes notation.

pydy_viz/tests/test_visualization_frame_scene.py
tarzzz, line 172: Once We have made relevant changes in the VFrame class and Scene, as per discussion on github, I will merge 
everything and then modify tests according to overall modifications, since we are going to change init methods and stuff, Tests would require major changes.

pydy_viz/tests/test_visualization_frame_scene.py
tarzzz, line 275: We have near and far, but In Ortho camera, the fov concept is decided from window height, and width.
So I will directly utilize it from Scene.height and width in JS side.

pydy_viz/scene.py
tarzzz, line 234: actually this method should take json data directly from the self._saved_json_file, this argument is not required.
I will remove it

pydy_viz/server.py
tarzzz, line 6: The index.html would be generated in the directory from where interpreter is launched. hence server would also be created there, and viz would be directly available at http://localhost

moorepants, line 6: But do you need to generate an html file in the user's working directory? Can't this be a standard thing that is installed into the source tree and you just load it?

tarzzz, line 6: the server can serve files only from the working directory, we cant access the html file from the installation directory.
Moreover, the html file would have to be modified, to change the name of the json file, (users can define names for json files being written), which should be accomplished by re.subs()

moorepants, line 6: Seems like these kind of questions will be answered once you present a JS api.

Seems like we haven't thought of some things:
- Should the user ever see the index.html file in the their working directory? I would think we wouldn't want them to see that or even edit it.
- Should we go with the hackish re.subs() methods of injecting things into our html and js files? It doesn't seem like a good idea, in general.
- Should the use be able to select other json files to load once they have the web app running? (i.e. if I do scene.display() I'd except to see the scene that I just defined, but should I be allowed to select another json file in the web app or do I need to run the python command for that one?
- What happens when you have multiple json files in one working directory. These can be different sims of the same model or sims for different models. MGView's method is to not be tied to the server side program (in their case MotionGenesis). MGView works with any correctly formatted json files you provide it.

tarzzz:
Merging the PR, display methods and tests modifications would directly be merged in master.

PR #20, opened 2013-08-07
Javascript tests

tarzzz:


pydy_viz/static/js/spec/CanvasSpec.js
```diff
@@ -0,0 +1,27 @@
...
+    // from it.
+    var JSONObj = // What to do ?
+    var canvas = new Canvas(JSONObj);
+    
+  });
+
+  it("should be able to initialize", function() {
+    canvas.initialize();
+
+    expect(canvas.scene.axes).toEqual(some_axes_stuff);
```
tarzzz, line 17: @moorepants It should be like this only, right?

pydy_viz/static/js/SpecRunner.html
```diff
@@ -0,0 +1,53 @@
...
+<head>
+  <title>Jasmine Spec Runner</title>
+
+  <link rel="shortcut icon" type="image/png" href="lib/jasmine-1.3.1/jasmine_favicon.png">
+  <link rel="stylesheet" type="text/css" href="lib/jasmine-1.3.1/jasmine.css">
+  <script type="text/javascript" src="lib/jasmine-1.3.1/jasmine.js"></script>
+  <script type="text/javascript" src="lib/jasmine-1.3.1/jasmine-html.js"></script>
+
+  <!-- include source files here... -->
+  <script type="text/javascript" src="src/Player.js"></script>
```
tarzzz, line 13: These lines need to be modified here ..

pydy_viz/static/js/spec/CanvasSpec.js
moorepants, line 17: I guess, like I said in the chat. I don't know how jasmine works. You know just as good as me.

PR #21, opened 2013-08-09
Camera class implementation

tarzzz:


tarzzz:
It is an earlier implementation which was not put in a PR, It is now merged and minute changes would be done in Master, If required

PR #23, opened 2013-08-10
Fixed import error from server.py.

moorepants:


moorepants:
And whitespace it seems. My new vim setup must automatically get rid of whitespace.

PR #25, opened 2013-08-10
Scene class

tarzzz:


pydy_viz/scene.py
```diff
@@ -103,18 +104,22 @@ def __init_(self, reference_frame, origin, *visualization_frames, **kwargs):
         self._origin = origin
             
 
-        self.visualization_frames = [vis_frame for visframes in visualization_frames]
+        self.visualization_frames = [vis_frame for vis_frame \
+                                              in visualization_frames]
```
moorepants: Can't you just do this:

self.visualzation_frames = list(visualization_frames)

moorepants:
You should probably rebase this off of master so that travis works and the changes I made yesterday are in.

pydy_viz/scene.py
```diff
@@ -164,25 +174,32 @@ def _generate_data(self, dynamic_variables, constant_variables, \
...
+                                        
+            if isinstance(frame, VisualizationFrame):                             
+                self._scene_data['frames'].append( \
+                                    frame.generate_visualization_dict())
+            else:
+                self._scene_data['cameras'].append( \
+                                    frame.generate_visualization_dict())
+        return self._scene_data
+                                        
+    def generate_visualization_dict(self, dynamic_variables, \
```
moorepants, line 163: What is the point of this method? Seems like you generate the dictionary `self._scene_data` in the above function. And this just creates a copy in `self._data_dict`.

PR #26, opened 2013-08-10
Added scene tests

tarzzz:


moorepants:
Why not include this with your Scene PR?

Also you need to include the static directory or this will fail.

moorepants:
Also, please merge or rebase to master, so we can start using Travis.

pydy_viz/tests/test_visualization_frame_scene.py
```diff
@@ -339,30 +344,51 @@ def test_orthographic_camera(self):
...
-        assert self.scene.frames[2] is self.frame3
+        self.scene2.reference_frame = self.A
+        assert self.scene2.reference_frame == self.A
         
+    def test_scene_copy_static(self):
+        self.scene1._copy_static_dir()
+
+        assert os.path.exists(os.path.join(os.getcwd()), '.pydy_viz'))
+        assert os.path.exists(os.path.join(os.getcwd()), \
+                                                    '.pydy_viz/static'))
```
moorepants: it is best to do this:

``` python
os.join.path(os.getcwd(), '.pydy', 'static')
```

That will ensure that it works on all OSs.

Also you don't need the backslashes at the end of lines when you have an open parentheses.

moorepants:
Where is the test for index.html?

tarzzz:
I am not sure we want index.html. I am reading on sockets, so that we can implement socket framework.
Then we wont need index.html at all. I will read on it today, and then write the required tests for sockets etc.

tarzzz:
I am merging this and Scene class, so that we can start using travis.

moorepants:
Ok, great.

PR #27, opened 2013-08-11
Added new Socket Server initializer

tarzzz:


pydy_viz/camera.py
```diff
@@ -15,44 +15,44 @@ class PerspectiveCamera(VisualizationFrame):
     2)ReferenceFrame, Point
     3)ReferenceFrame, Particle
     Either one of these must be supplied during initialization
-
```
tarzzz, line 18: One command from bash, and all the white spaces are vanished within a second. 
I <3 bash.

PR #30, opened 2013-08-18
Javascript tests

tarzzz:


moorepants:
These look fine but only test for the attributes that are set.

What about the methods? What methods will your class have? These need to be tested too.

pydy_viz/static/js/spec/CanvasSpec.js
```diff
@@ -1,27 +1,93 @@
...
-  var canvas;
-  
-
-  beforeEach(function() {
-    // basic functionality test is to describe some standard JSONObj
-    // here and then simply check the different methods for the values
-    // from it.
-    var JSONObj = // What to do ?
+describe("Canvas Class for Animation", function() {
     var canvas = new Canvas(JSONObj);
```
tarzzz, line 2: We test constructor from here ..

pydy_viz/static/js/spec/CanvasSpec.js
```diff
@@ -1,27 +1,93 @@
...
+    expect(canvas.height).toEqual(JSONObj.height); });
+
+    it("should have a predefined Object3D for adding axes", function() {
+    expect(canvas.axes instanceof THREE.Object3D).toBeTruthy(); });
+
+    it("should have a mesh dictionary for holding shape objects", function() {
+    expect(canvas.mesh_dict).toBeDefined(); });    
+
+    // Now we initialize the canvas ...
     canvas.initialize();
```
tarzzz, line 22: We initialize the canvas. By initializing I mean adding dynamic button/divs, and adding axes and stuff. They are now added in the 
class object. So we can test it in below lines

pydy_viz/static/js/spec/CanvasSpec.js
```diff
@@ -1,27 +1,93 @@
...
 
+    it("should have a add_frames" function() {
+    expect(canvas.add_frames).toBeDefined(); });       
+    
+    canvas.add_frames();
+    
+    it("should have a add_frames" function() {
+    expect(canvas.add_frames).toBeDefined(); });       
+    
+    s.add_frames();
```
tarzzz, line 67: Once canvas is initialized fully(buttons etc. are loaded) Instruct the object to add vframe shape data to the scene, using this method
Once added, we can check whether they are actually added as follows

pydy_viz/static/js/spec/CanvasSpec.js
```diff
@@ -1,27 +1,93 @@
...
+    
+    it("mesh dictionary member 2 should match from JSON Object." function() {
+    expect(canvas.mesh_dict['frame2'] instanceof THREE.Mesh).toBeTruthy(); });       
+
+    it("mesh dictionary member 3 should match from JSON Object." function() {
+    expect(canvas.mesh_dict['frame3'] instanceof THREE.Mesh).toBeTruthy(); });       
+    
+    // Add TrackBall controls should be defined ... 
+    it("should have a  method to initiate trackball controls" function() {
+    expect(canvas.add_track_ball_controls).toBeDefined(); });
```
tarzzz, line 85: So I cant call add_track_ball_controls method since It will take the system into infinite loop, since we are using requestAnimationFrame here, Hence I am checking only whether It is defined in the object.
Please Suggest something what I can do to check here.

pydy_viz/static/js/spec/CanvasSpec.js
```diff
@@ -1,27 +1,93 @@
...
+    expect(canvas.mesh_dict['frame2'] instanceof THREE.Mesh).toBeTruthy(); });       
+
+    it("mesh dictionary member 3 should match from JSON Object." function() {
+    expect(canvas.mesh_dict['frame3'] instanceof THREE.Mesh).toBeTruthy(); });       
+    
+    // Add TrackBall controls should be defined ... 
+    it("should have a  method to initiate trackball controls" function() {
+    expect(canvas.add_track_ball_controls).toBeDefined(); });                
+    
+    //TODO Add Check for run animation once ..
```
tarzzz: I talked with people on Irc on this one. am working on same.

PR #35, opened 2013-08-28
Light class

tarzzz:


pydy_viz/__init__.py
```diff
@@ -28,6 +28,11 @@
 __all__.extend(camera.__all__)
 
 
+import light
+from light import *
```
moorepants, line 32: why have two imports here?

moorepants:
This is generally fine too. If you make your PR with respect to the branch you branched from (like the light test branch) then this would only show the commits from this branch.

PR #37, opened 2013-09-04
Roll back js travis

tarzzz:


moorepants:
What is the deal with the failure here? You reshaping seems to be messed up.

Nice avatar.

tarzzz:
I had coreected it. But I forgot in which branch.
I will correct it in master directly.

tarzzz:
And btw that is some default avatar, coz I dont remember puttig up one .. :)

PR #38, opened 2013-09-09
Finished basic pydy viz

tarzzz:


moorepants:
I thought we talked about three atomic pull requests:
1. Python code for Light Class and the accompanying tests.
2. Canvas class in javascript with accompanying tests.
3. JS test integration with command line testing and Travis support.

This PR looks like it has elements off all of that in it.

tarzzz:
I made PR to run travis tests.
It was only for that purpose.
I am going to make atomic PR's, for discussion purposes.

moorepants:
The point of atomic PRs isn't just for discussion purposes. It allows us to have smaller chunks of independent functionality code that we can review and merge with working tests. A PR with 33 files changed and 17 thousand additions is extremely hard to review. Furthermore, if we ever have to revert changes it is impossible if you have unrelated changes in the same commits.

pydy_viz/scene.py
```diff
@@ -216,9 +216,9 @@ def generate_visualization_dict(self, dynamic_variables,
         self._scene_data['width'] = self._width
         self._scene_data['frames'] = []
         self._scene_data['cameras'] = []
+        self._scene_data['lights'] = []
 
-        for frame in self.visualization_frames + self.cameras:
-
+        for frame in self.visualization_frames:
             frame.generate_transformation_matrix( \
```
catchmrbharath, line 222: You dont require the `\` here. You can insert new lines within brackets without escaping,

moorepants, line 222: Correct. Tarun, anytime you have parentheses, brackets, etc with the opening on one line and the closing on another, Python does not need the backslash.

pydy_viz/light.py
```diff
@@ -0,0 +1,176 @@
...
+        >>> rbody = RigidBody('rbody', O, I, mass, (inertia, O))
+        >>> # Initializing with a rigidbody ..
+        >>> light = PointLight('frame2', rbody)
+        >>> Pa = Particle('Pa', O, mass)
+        >>> #initializing with Particle, reference_frame ...
+        >>> light = PointLight('frame3', I, Pa)
+        """
+        
+        try:
+            self._color = kwargs['color']
```
catchmrbharath, line 74: You can use the `get` method here, instead of catching exception. Something like `kwargs.get('color', default='white')`.

moorepants, line 74: That's a nice one. @tarzzz you should use that to clean up much of the code.

Issue #39, opened 2013-09-09
Passing in a particle to Visualization frame raises an error


moorepants:
If I do this:

VisualizationFrame('test', particle, shape)

I get a "particle doesn't have a dcm error"

This should be caught and say "You need to pass in a reference frame and a particle!"

moorepants:
Fixed in PR #49

PR #43, opened 2013-09-10
Mime type and raise Error for Particle arg

tarzzz:
Added content header for fixing raw html problem in firefox.
raised Error if only Particle/Point is supplied w/o ReferenceFrame in VisualizationFrame, Light and Camera
Tests are failing, but they will pass when Javascripts are merged.


moorepants:
These should be separate PRs since they are completely unrelated and independent.

You can also reference issue numbers in the commit message or github comments like this: issue #39 so we can reference which commit (or PR) deals with what issue.

PR #46, opened 2013-09-11
Javascripts

tarzzz:


pydy_viz/static/js/canvas/canvas.js
```diff
@@ -0,0 +1,195 @@
...
+        console.log(_frame.shape);
+        switch(_frame.shape.type){
+        
+        case "Cylinder":
+            var color = new THREE.Color().setRGB(_frame.shape.color[0],
+                           _frame.shape.color[1], _frame.shape.color[2])
+                                    
+            var _material = new THREE.MeshLambertMaterial({
+                                     color:       _frame.shape.color,
+                                     wireframe:   true,
```
moorepants: I think the default for this should be false. The wire frames don't look good.

pydy_viz/static/js/canvas/canvas.js
```diff
@@ -0,0 +1,195 @@
...
+        this.frames.add(_mesh);
+       
+    }
+    this.scene.add(this.frames);
+};
+
+
+
+Canvas.prototype.startAnimation = function(){
+    for(var i = 0; i< 100; i++){
```
moorepants: `100` need to be generated from the number of frames that are in the json object. Otherwise this fails for different frame length numbers.

moorepants:
What is the difference in this PR and PR #38 ?

pydy_viz/static/js/canvas/canvas.js
```diff
@@ -0,0 +1,195 @@
...
+        // case "MoreLights" ....
+        
+        }
+        this.lights.add(_Light);
+       
+    }
+    this.scene.add(this.lights);
+};
+
+Canvas.prototype.addFrames = function(){
```
moorepants: Why does this exist in this file and in addObjects.js?

pydy_viz/static/js/canvas/canvas.js
```diff
@@ -0,0 +1,195 @@
...
+        }
+        this.frames.add(_mesh);
+       
+    }
+    this.scene.add(this.frames);
+};
+
+
+
+Canvas.prototype.startAnimation = function(){
```
moorepants: Why does this also exists in addObjects.js?

pydy_viz/static/js/lib/TrackballControls.js
```diff
@@ -0,0 +1,537 @@
+/**
```
moorepants, line 1: For all these js libs that we don't actually edit, wouldn't it be better to just include the minified versions?

pydy_viz/static/js/canvas/addObjects.js
```diff
@@ -0,0 +1,200 @@
...
+                                    
+        var _material = new THREE.MeshLambertMaterial({
+                                    color:        _frame.shape.color,
+                                    opacity: 0.5
+                                                     });        
+
+        switch(_frame.shape.type){
+
+        case "Cube":
+            alert("In Cube");
```
moorepants: remove

pydy_viz/static/js/canvas/addObjects.js
```diff
@@ -0,0 +1,200 @@
...
+                                      _frame.shape.radius,
+                                            _frame.shape.radius,
+                                      _frame.shape.height,
+                                      50,50);
+            break;
+                                      
+        case "Cone":        
+            var _geometry = new THREE.CylinderGeometry(
+                                      _frame.shape.radius,
+                                      0.1,
```
moorepants: Why do we have a separate Cone class? Seems like you should just have the Cylinder class in python and allow the user to set both radii.

pydy_viz/static/js/canvas/initialize.js
```diff
@@ -0,0 +1,113 @@
...
+	
+    this.grid.add(gridYZ);
+    this.scene.add(this.grid);
+    
+    this.axes = new THREE.AxisHelper(JSONObj.height);
+    this.axes.name = "Axes";
+    this.scene.add(this.axes);
+    this.renderer.render(this.scene, this.primaryCamera);
+    
+    // copying to vars for
```
moorepants: ?

pydy_viz/static/js/canvas/initialize.js
```diff
@@ -0,0 +1,113 @@
...
+    this.renderer.render(this.scene, this.primaryCamera);
+    
+    // copying to vars for 
+    
+    // Setting a basic Point Light for colors ...
+    this.primaryLight = new THREE.PointLight(0xffffff);
+    this.primaryLight.position.set(10,10,-10);
+    this.scene.add(this.primaryLight);
+    // A point object to show this light ..
+    var _geom = new THREE.SphereGeometry(2,100,100);
```
moorepants, line 96: These glowing spheres for lights should be optional. I can imagine using them to help setup the scene, but not beyond that.

pydy_viz/static/js/lib/TrackballControls.js
moorepants, line 1: Is it a big deal to include the minified versions of three.js, trackballs, and jquery?

pydy_viz/static/js/canvas/addObjects.js
tarzzz: I feel its better to keep them different. It keeps it simple for user to
use.

moorepants:
This is currently failing on travis because of a missing canvas.js file.

tarzzz:
We need to merge master to this branch here, and it will settle.

moorepants:
I just ran the example conical pendulum using this branch. The computations completed, but firefox shows the raw html file and chrome shows a few buttons, no visualization. Seems like something is missing...

pydy_viz/static/js/canvas/addObjects.js
moorepants: We can keep the Cone class but what is some makes a model of a tiny cone and they set the cone base radius to be 0.01, what happens then?

I think you should allow an optional argument on the cylinder to set the second radius (top or bottom) for a case like I just mentioned. Setting the radius to 0.1 is problematic. You should at least move that to an optional arg in the Cone class on the python side.

moorepants:
This still doesn't work for me. What branches do I need to merge to get this working? If I can see it working, I'll give a thumbs up for merging.

pydy_viz/static/js/canvas/addObjects.js
```diff
@@ -0,0 +1,200 @@
...
+                                      _frame.shape.length,
+                                      _frame.shape.length,                                                                            
+                                      50, 50, 50);          
+            break;                          
+
+        case "Cylinder":
+            var _geometry = new THREE.CylinderGeometry(
+                                      _frame.shape.radius,
+                                            _frame.shape.radius,
+                                      _frame.shape.height,
```
moorepants: I still don't think this is correct. Should it be _frame.shape.length?

You need some tests that ensure you have the right variable names here. I think on the python side you call the a length, not a height. Why doesn't an error raise if there is no height attribute?

The cylinders in the example are still too long!

moorepants:
I merged master in and it works except for the lengths of the cylinders. Did not not merge my pull request with fixes to this?

moorepants:
It works in firefox after merging master into this branch too.

pydy_viz/static/js/lib/TrackballControls.js
moorepants, line 1: Are you going to respond to this?

pydy_viz/static/js/canvas/addObjects.js
moorepants: Still needs to be removed. Why is there an alert?

pydy_viz/static/js/canvas/addObjects.js
moorepants: I like the idea of a proportional tip radius. Please make that change.

pydy_viz/static/js/canvas/initialize.js
moorepants: Still here.

pydy_viz/static/js/lib/TrackballControls.js
tarzzz, line 1: Oh  yes, three and jquery are already minified. I need to change their extensions. As for TrackballControls and jquery-ui.js they are not available in minified versions, So I have kept them as such.

moorepants, line 1: Ok, please fix the extensions. We can minify the other ourselves later.

moorepants:
Ok, this branch works correctly for me now, at least for the conical pendulum example (after some things are fixed in the example code). So it is basically ready to go in. I still need to run the tests though.

There is one bug though: issue #71 . What would it take to fix this bug?

tarzzz:
I am working on it right now.

tarzzz:
Should me merged in master, closing

PR #48, opened 2013-09-12
Sphinx doc

tarzzz:


docs/src/installation.rst
```diff
@@ -0,0 +1,84 @@
...
+
+
+There are different methods you can use to get PyDyViz up and running for your system.
+
+Dependencies:
+
+1) Python >=2.5
+2) SymPy >=0.7.2
+3) NumPy >=
+4) SciPy >=
```
moorepants: Just write NumPy and SciPy with no ">=", I don't think we use any functionality from those packages that is brand new. It will probably work with versions that go pretty far back.

You should include matplotlib too. Right now it depends on that for the color module right?

docs/src/installation.rst
```diff
@@ -0,0 +1,84 @@
...
+
+You can check the version of the software by issuing following command from interpreter::
+
+    >>> import pydy_viz
+    >>> print pydy_viz.__version__
+
+Questions
+=========
+
+If you have any question about installation, or any general question, feel free to visit the IRC channel at irc.freenode.net, channel #pydy.
```
moorepants: wrap these lines at 79 characters

moorepants:
The makefile that sphinx autogenerates doesn't seem to be in the git repo. Please at the Makefile.

docs/src/installation.rst
tarzzz: No we dont need matplotlib. actually I figured out THREE.js can accept color strings directly for its input. 
So no need to convert to rgb or any other values. We can remove matplotlib's color converter altogether.
I can do it in master once we have other things got merged in.

tarzzz:
Could not find any autogenerated Makefile in the repo.

moorepants:
This is a standard Makefile generated by sphinx-quickstart: https://github.com/chrisdembia/yeadon/blob/master/doc/Makefile

docs/src/installation.rst
moorepants: Can the user of pydy-viz pass in either color strings or rgb values if we remove matplotlib requirements? We need a simple way to set colors (names) and a way that allows the user to specify any color they want (rgb, hex, or whatever).

moorepants: Also, I don't think that scipy is a requirement for pydy-viz. We don't integrate any equations in this package. We only use numpy for storing arrays and lambdify for creating numerical functions from sympy expressions, right?

tarzzz: We arent using scipy in pydy_viz specifically.
Also as for matplotlib is concerned, direct color strings is the easy way to add colors. and three has support for hex and rgb alike. We can add them as a later feature. Right now I think simple color strings without matplotlib are good.
I dont think we would require matplotlib's color converter, even when we add hex and rgb features.

moorepants: Ok, so let's remove this scipy requirement. Check the setup.py file too and make sure scipy is not there.

If only color strings work now, that is fine for the first release. Please remove any mention of matplotlib in the code if we are not using it (especially the travis tests and setup.py).

docs/src/installation.rst
```diff
@@ -0,0 +1,94 @@
...
+
+There are different methods you can use to get PyDyViz up and running 
+for your system.
+
+Dependencies:
+
+1) Python 
+2) SymPy 
+3) NumPy 
+4) SciPy
```
moorepants: The code generator does not exist in this package, so I think you should remove the SciPy dependency.

docs/src/installation.rst
```diff
@@ -0,0 +1,94 @@
...
+
+If it does not throws any error/traceback, it means that PyDyViz 
+is installed.    
+
+
+You can check the version of the software by issuing following 
+command from interpreter::
+
+    >>> import pydy_viz
+    >>> print pydy_viz.__version__
```
moorepants, line 53: Do you actually set the version in the source code?

tarzzz:
Need to add JS Docstrings, then good to merge.

docs/src/installation.rst
moorepants: This still needs to be removed.

docs/src/javascript_api.rst
```diff
@@ -0,0 +1,99 @@
...
+  - A Div for displaying total number of frames.
+  - A Div for displaying the current frame animation is
+    running on.
+
+canvas/addObjects.js
+====================
+Canvas.prototype.addControls
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This prototype function initializes the Primary Controls,
```
moorepants, line 41: Is there no way to load these into the Sphinx docs automatically so you don't have two places that you have to manually keep in sync?

I sent you a link to a sphinx extension for js a while back that seemed to do that.

tarzzz, line 41: Yeah I saw the link pyjsrest.
It seems to be part of some package. Moreover it wasnt MIT licence. So I
havent included it for now.

moorepants, line 41: Ok, this is fine for now but please make an issue that says we need to automate the js doc strings extraction for the sphinx docs.

tarzzz, line 41: Sure, I will. We can merge it if you have gone through once.Its complete in
itself.

moorepants:
You still need to address the scipy note i left and the **version** variable. Are these fixed?

tarzzz:
Yes they are fixed!

moorepants:
But I still see Scipy listed as a dependency list.

docs/src/installation.rst
moorepants: SciPy is not a dependency for this package, so why do we include it?

tarzzz:
Ohh my bad. I thought I had removed it.

moorepants:
once you fix that it is good to merge.

tarzzz:
I am closing this PR, so that it can be merged in master.

PR #50, opened 2013-09-12
Default light

tarzzz:


moorepants:
This PR seems fine. The only addition could be adding a test to check that this kwarg was set correctly. Is it tested?

tarzzz:
Sorry, I didnt get you ..

moorepants:
Sorry you didn't get me?

tarzzz:
What test are you rwferring to.i mean I have pointlight tests.

moorepants:
Here you add a new kwarg to Scene (i think). Do you test this functionality?

tarzzz:
@PythonDynamics/visualization I havent tested this functionality. i am going to create another PR for that.
Merged this one.

moorepants:
Please don't create new PRs for tests that should belong with this function in the future. The reason is that you just merged untested code into master. What if this had a bug in it? Always merge "tested" code into master unless you really just don't plan to ever write a test for it (which ideally would never be the case).

PR #58, opened 2013-09-17
Js docstrings

tarzzz:


moorepants:
Travis is actually failing: https://travis-ci.org/PythonDynamics/pydy-viz/builds/11450950

The bin/test command is exiting with zero even though the tests fail.

moorepants:
I didn't look at all the details of this, but this seems complete in itself, even if you don't have a docstring for everything. This looks fine to merge.

tarzzz:
The ts failing only because master isnt merged in it.  I will do that and
then it will pass

moorepants:
These three commits don't affect any tests that are run on travis, so it can be merged as is.

PR #59, opened 2013-09-17
Js testing

tarzzz:


pydy_viz/static/js/jasmine/jasmine-html.js
```diff
@@ -0,0 +1,681 @@
+jasmine.HtmlReporterHelpers = {};
```
moorepants, line 1: Why inlcude the jasmine source when you can install it with npm? Should you just have install instructions for this dependencie with npm instead of including it in our source?

moorepants:
How do I run this? Where are the instructions for running these tests?

tarzzz:
Its in master README

moorepants:
ok, why do you need to run a python web server to run the js tests?

tarzzz:
yeah actually some issue in blanket.js, which we use for code coverage. It
doesnot allow cross domain requests, hence gives error on loading local
files. So a simple server is required.

moorepants:
Ok.

pydy_viz/static/js/jasmine/jasmine-html.js
tarzzz, line 1: We wont need it when we automate the testing. But for now, We need to open an html page, and hence it requires source. It is a temporary dependency, until we automate JS testing

moorepants:
I just ran the tests and my whole computer locked up...not sure what that was all about.

tarzzz:
That is the the webgl problem.
My system itself slows down during js tests. Need to look into it

tarzzz:
I thought it was some system heating problem. But if your system is also
having the same thing I might need to look into it.

moorepants:
I'll try running the tests again and see if it works.

moorepants:
The tests cause both Firefox and Chrome to lock up for me. It also cause everything else to lock up. My mouse moves but doesn't seem to affect anything. I waited a long time for the Chrome lockup and kept pressing alt-F4 and eventually the window closed and my computer was back to normal. It seems like something serious is hanging the tests, at least on my machine. We can't merge this until we figure it out.

pydy_viz/static/js/jasmine/jasmine-html.js
moorepants, line 1: ok

moorepants:
Ok, 43 test pass for me with no crashes now. The coverage is low, is that because or the most recent changes or is that just how it is?

This looks good to merge, but if there is work to be done to get it to run properly still, then make an issue for it.

tarzzz:
Except code coverage issues, rest it works fine!. 
I am merging it!

PR #72, opened 2013-09-19
Clean server shutdown

tarzzz:


moorepants:
Why are all these extra commits in this PR, when you should only have the last one?

tarzzz:
Addressed in https://github.com/PythonDynamics/pydy-viz/pull/75

PR #93, opened 2014-01-25
Minimize js canvas

tarzzz:
Minimized version of canvas module is helpful for:
1. Optimized rendering.
2. would be very helpful for IPython plugin


moorepants:
Is canvas.js the javascript you wrote? I forget.

Some notes on ?good? practices for js dependencies:
- Right now we include js dependencies for three.js, jquery.js, etc. These can  be committed to the git repo as minified files, as we don't change them, except on version bumps. We just use them as libraries.
- The js files we write should be committed to repo as un-minified. We could also include a minified version, but ideally the minification should happen on installation of pydy-viz or when we create a source distribution to upload to PyPi.

Also keep in mind that minification doesn't increase execution time, it is there for two reasons: smaller downloads and obfuscation.

The download issue is important for websites because users have to download the js files every time they visit a website (except for caching), but this issue doesn't really matter for us because the user gets the js libraries on their first download of the whole package and it is served up locally.

tarzzz:
I guess we can leave the scripts unminified. As it is not required at the moment

PR #94, opened 2014-01-25
Removed print statements, Added Mime type for proper firefox rendering

tarzzz:


moorepants:
Shouldn't this just merge to master?

moorepants:
I confirmed that this works on firefox. Good to merge.

moorepants:
What is the difference in this and PR #98?

tarzzz:
My bad! I have merged the other one. Both are exactly same!

PR #100, opened 2014-01-27
pep8 improvements and README correction

pramttl:
Added pep8 compliance to some files without changing functionality.There were lot of pep8 problems that make the code visually hard to understand and non compliant to pep8 standards.

Here is a nice python tool that is useful to check pep8 compliance: [pep8](https://pypi.python.org/pypi/pep8)


moorepants:
Thanks.

PR #101, opened 2014-01-29
Modified scene.py to run server in background, when called from IPython notebook

tarzzz:
This is a hack which can serve the purpose (atleast until the conferences). It would not be required once IPython2.0 is out, as we would be serving visualizations in output cells, and not from a separate server

### DIFF
--- a/pydy_viz/server.py
+++ b/pydy_viz/server.py
@@ -98,7 +98,6 @@ class Server(threading.Thread):
                 self.data = conn.recv(1024)
                 sent_data = self._parse_data(self.data)
                 conn.send(sent_data)
-                print 'sent data'
                 conn.close()
             except KeyboardInterrupt:        
                 print "Are you sure you want to shutdown[Y/N]?"
--- a/pydy_viz/server.py
+++ b/pydy_viz/server.py
@@ -49,7 +49,13 @@ class Server(threading.Thread):
 
         static_path = os.path.dirname(pydy_viz.__file__)
         static_path = os.path.join(static_path, 'static')
-        request = data.split(' ')[1]
+        try:
+            request = data.split(' ')[1]
+        except KeyError:
+			#If error occurs in parsing a request,
+			#Better to reload the page, to avoid broken
+			#javascripts
+			request = '/'    
         if request == '/':
         #If requested for http://localhost:port/
         #Send index.html file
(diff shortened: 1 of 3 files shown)


tarzzz:
@moorepants Please review it, so that It can be merged

moorepants:
This seems to work, some notes:
- when you type scene.display() no text shows up in the output cell in the notebook until you actually navigate to the localhost:8000 url. But you need to know the url before you can navigate there.
- This branch does not work in firefox, seems like the same issue as before, i.e. just shows the raw html (note that the MSAB tutorial may or may not have chrome).
- The new close server button should be on a new row, it didn't fit in my window and it gets chopped in half when it wraps to a new row.

moorepants:
Here is the weird button:

my first three js app - chromium_007

moorepants:
Notice that if scene.display() wasn't blocking you could likely do this to get the output in an IPython cell:

selection_006

Note that the text that gives the url for the viz doesn't show up until after you load the url in your web browser.

tarzzz:
1. Weird button: I am not concentrating on UI( I am going to change UI into a decent one, once features are working).
2. scene.display() no text: This is how it looks from my side(maybe you didnt checked out on this branch before running or correctly set PYTHONPATH). It should show a new thread being started in background.
   selection_001
3.  After I call HTML("localhost:8000"):
   selection_002

tarzzz:
Firefox Support was merged in master after this branch split. So I have merged master into it.

moorepants:
Ok, I'll try again tomorrow. Going to sleep now. The weird button can be fixed by a simple `<br \>` tag to put it on the next line.

pydy_viz/server.py
```diff
@@ -61,6 +61,11 @@ def _parse_data(self, data):
             file_path = os.path.join(os.getcwd(), self.saved_json_file)
             send_buffer = 'var JSONObj = '
 
+        elif request == '/close-server':
```
moorepants, line 70: i think this needs to be:

elif request == r'/close-server`

or

'\/close-server'

This is the url I'm  getting:

http://localhost:8001/\close-server

moorepants:
The visualization looks decent now in the master branch of the tutorial.

More notes:

Getting this error
n08_visualization - mozilla firefox_010

Is there any way that scene.display() is not blocking or that it at least displays the text correct?

busy n08_visualization - mozilla firefox_009

pydy_viz/server.py
tarzzz, line 70: Link was incorrect in index.html.
Is fixed now.. pushing code

tarzzz, line 70: Display doesnt show busy notebook on my side.
That error I need to look into.
I have same sometimes.

moorepants:
If you think this all good, you can merge. It's better than it was, for sure.

PR #103, opened 2014-02-12
[WIP] Improved ui

tarzzz:
Working on a better UI.


moorepants:
Some comments:
- The colors look more washed out in the 3D portion. Before the colors were more crisp.
- I don't think we need to show a sphere representing the light. It is confusing to know what that is.
- I like the new nice looking buttons.

pydy_viz/static/index.html
```diff
@@ -1,54 +1,136 @@
...
-        <link rel="stylesheet" type="text/css" href="css/style.css">
-		<script type="text/javascript">
-		var alert1 = function () { alert("1"); }
-        $( "#slider" ).slider();
-	</script>	      
-	</head>
 
+		<!--- Bootstrap --->
+		
+		<script src="js/lib/bootstrap.min.js"></script>
```
moorepants, line 16: Ahh, that's where you got the nice buttons.

pydy_viz/static/js/canvas/initialize.js
```diff
@@ -60,7 +60,7 @@ Canvas.prototype.initialize = function(){
 
 	this.renderer = new THREE.WebGLRenderer();
 	this.renderer.setSize(JSONObj.width, JSONObj.height);
-	
+        this.renderer.setClearColor(0xcccccc);
```
moorepants: Is this the color change?

moorepants:
Also, why doesn't the 3D portion fit in the browser window. On my machine it is longer than the browser is tall. Is there a way to scale the height of the 3D view to the height of the browser window. Then you can make the visualization some percentage of the width. Btw, you can probably decrease the amount of screen real estate taken up by the control panel.

moorepants:
I just tried this on the tutorial problem. The overall rendering speeds seems slower than on master.

PR #106, opened 2014-02-18
Added matplotlib dependency specification

tarzzz:


moorepants:
Added some updates here and merging.

PR #110, opened 2014-02-22
Added mock to docs/conf.py to allow building on readthedocs.

moorepants:


moorepants:
This makes use of readthedocs virtualenv and --system-site-packages (numpy and matplotlib are installed on rtd machines). Might be able to remove the mock code if the libraries are actually there. I had lots of trouble because RTDs installs your package and to do that it needs all the dependencies, of which numpy and matplotlib are a pain to install.

moorepants:
Docs are here: http://pydy-viz.readthedocs.org

moorepants:
@tarzzz Make an account on RTDs and I'll give you access to the documentation building, etc.

tarzzz:
Sure.  I will .
Name should be pydy-viz or pydyviz?

moorepants:
I just decided to stick with "pydy", "pydy-code-gen", and "pydy-viz" instead of changing things.

Eventually, I think we will merge everything into "pydy" and only have one package. Maybe we can do that in Montreal.

tarzzz:
Cool! pydy-viz it is..
I will create an account on RTDs and will mail you the credentials/add you.

PR #117, opened 2014-03-30
Added create_static method to Scene class. 

tarzzz:
It allows you to create static webpages for the visualization.
Added cleanup_static method to remove the static data generated!

**To test this branch**
modify last line in <code>pydy_examples/illustrative_example_for_viz/illustrative_example.py</code>
from:
<code> scene.display()</code>
to:
<code> scene.create_static()</code>
and run it:
<code> python illustrative_example.py</code>

then cd to static directory generated, and from terminal
<code>$ python -m SimpleHTTPServer</code>
You should be able to see visualization at localhost:8000

### DIFF
--- a/pydy_viz/scene.py
+++ b/pydy_viz/scene.py
@@ -316,24 +316,45 @@ class Scene(object):
                                  separators=(',', ': ')))
         outfile.close()
 
-    def _copy_static_dir(self):
+    def create_static(self):
         """
         Copies all the static files required in the
         visualization to the current working directory
         The files and sub directories are stored within
-        a hidden directory named .pydy_viz in the current
+        a directory named static in the current
         working directory.
-        Working directory can be cleaned by calling _cleanup()
-        method, which deletes the .pydy_viz directory.
+   
+        This method is used to output static files for embedding
+        the visualizations in the static webpages. Simply copy 
+        the contents of static directory in the relevant directory
+        for embedding in a static website.
 
         """
-        dst = os.path.join(os.getcwd(), '.pydy_viz')
-        os.mkdir(dst)
-        src = os.path.join(os.path.dirname(pydy_viz.__file__), 'static')
-        distutils.dir_util.copy_tree(src, dst)
+        dst = os.path.join(os.getcwd(), 'static')
+        try:
+            os.mkdir(dst) 
+        except OSError:
+            print 'Directory already exists, refreshing contents..'
 
-    def _cleanup(self):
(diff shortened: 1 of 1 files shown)


tarzzz:
@moorepants Please review and merge it. 
P.S.: it was already partially implemented in scene class, but was redundant earlier. Now we can have a better use of it :)

moorepants:
This works. No need to open a server. Simply open index.html in your browser.

Gonna add a couple of things before merging.

moorepants:
These really need tests before merging. Not sure how long that would take. @tarzzz, are you fine with the changes I made.

pydy_viz/scene.py
```diff
@@ -316,24 +317,84 @@ def generate_visualization_json(self, dynamic_variables,
...
-        dst = os.path.join(os.getcwd(), '.pydy_viz')
-        os.mkdir(dst)
-        src = os.path.join(os.path.dirname(pydy_viz.__file__), 'static')
-        distutils.dir_util.copy_tree(src, dst)
 
-    def _cleanup(self):
-        distutils.dir_util.remove_tree(os.path.join(os.getcwd(), '.pydy_viz'))
+        dst = os.path.join(os.getcwd(), 'static')
+
+        if os.path.exists(dst) and overwrite is False:
```
tarzzz, line 342: I am fine with the changes, but I feel we should be writing tests too for the same(because of conditional statements). I can write them and if we are short of time(sending tutorial instructions to students), then we can merge it and tests can be sent in a separate PR.

moorepants, line 342: This isn't critical to have for the tutorial, because we already have at least one browser/os combination that works for everyone. But if this was in a new pydy-viz release before the april 2, then we would have it available for backup in case the socket server isn't working for some.

moorepants:
@tarzzz You need to get this merged into the pydy repo.

tarzzz:
I am fine with the changes, but I feel we should be writing tests too for
the same(lots of conditional statements). I can write them and if we are
short of time(sending tutorial instructions to students), then we can merge
it and tests can be sent in a separate PR.

tarzzz:
Sorry, that was a draft I made some time ago(before PyCon). I am onto it
now.

moorepants:
@tarzzz 

This would be great to have working in the scipy tutorial. Can you get this working in the new pydy repo?

We plan to send out final instructions by June 23rd so it needs to be done sometime before that.

PR #118, opened 2014-03-30
Added MIME-types to all hyperlinks in index.html

tarzzz:
Not Sure whether this will  fix the compatibility on Mac OS X Chrome, but it is a standard to add MIME types with hyperlinks.


moorepants:
This doesn't solve it. It must be that the socket server is sending the HTTP response header from the GET request at `text/plain`. We need to figure out how to specify this in the Python server code. But this should be merged regardless, as this is correct and preferred.

